Issue #50, opened 2025-04-20
Use Awaitility for Asynchronous Tests


eschizoid:

eschizoid:
https://github.com/awaitility/awaitility

PR #51, opened 2025-04-22
Implement Offset Manager

eschizoid:


eschizoid:
@rkolesnev Let me know if you can give it a quick look to this bad boy if you have time :)

Thanks in advance!

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,574 @@
...
+      initializePartitionOffset(partition);
+    }
+  }
+
+  /** Initialize the offset position for a partition. */
+  private void initializePartitionOffset(final TopicPartition partition) {
+    enqueueCommand(
+      () -> {
+        try {
+          nextOffsetsToCommit.put(partition, consumer.position(partition));
```
rkolesnev, line 180: This seems wrong - consumer.position(partition) returns next fetch offset. 
From javadoc:
Get the offset of the next record that will be fetched (if a record with that offset exists).
So initializing next offset to commit as offset of next poll start seems not right - should probably initialize on first poll to first offset polled... (and re-initialize on partition assignment).

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,574 @@
...
+    );
+  }
+
+  /**
+   * Marks an offset as successfully processed.
+   *
+   * @param record The consumer record that was processed
+   */
+  public void markOffsetProcessed(final ConsumerRecord<K, V> record) {
+    markOffsetProcessed(new TopicPartition(record.topic(), record.partition()), record.offset() + 1);
```
rkolesnev, line 195: i understand that offset to commit is processed offset + 1 - but i would move the +1 logic into the commit block. As it is not really semantically matching the tracking - here the method seems like its tracking offset that was processed - which is not offset +1 in itself.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,574 @@
...
+
+    // Find the highest contiguous processed offset
+    final var highestContiguous = findHighestContiguousOffset(processed);
+    final var nextExpected = highestContiguous + 1;
+
+    // If all offsets are processed, commit one past the highest contiguous
+    if (pending == null || pending.isEmpty()) {
+      nextOffsetsToCommit.put(partition, nextExpected);
+      return;
+    }
```
rkolesnev, line 383: this seems wrong? the processed offsets are already holding offset +1 as values - next expected is +1 again.
So lets say we processed offsets 0-9 inclusive - what actually is stored in processedOffsets is 1-10 inclusive and then nextExpected=10+1 - so 11... but it should commit offset 10 (last processed offset of 9 +1 ).

rkolesnev: I am not sure why having pending offsets has an impact here? 
Lets say offsets polled and submitted are 0-9 inclusive
processed 0-5, 
pending 6,7,
processed 8,9
logic should just commit 6 (as highest consecutive processed of 5 +1 ) regardless of pending offsets?

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,574 @@
...
+  private static final Logger LOGGER = Logger.getLogger(OffsetManager.class.getName());
+  private static final long DEFAULT_COMMAND_TIMEOUT_MS = 5000;
+
+  private final Consumer<K, V> consumer;
+  private final ExecutorService executorService;
+  private final AtomicReference<OffsetState> state = new AtomicReference<>(OffsetState.CREATED);
+
+  // Tracking data structures
+  private final Map<TopicPartition, ConcurrentSkipListSet<Long>> pendingOffsets = new ConcurrentHashMap<>();
+  private final Map<TopicPartition, ConcurrentSkipListSet<Long>> processedOffsets = new ConcurrentHashMap<>();
```
rkolesnev, line 50: I am not really convinced that both maps are needed. 

I think it would be a bit simpler to just maintain pending (inflight) offsets and nextOffsetsToCommit map.

On processing completion for the offset:
- remove from pending offsets
- set nextOffsetsToCommit if processed offset is above current value in nextOffsetsToCommit (would rename to highestProcessedOffset)

On commit - just commit up to Min(lowest pending, highestProcessedOffset +1 ) - so if there are no pending - just commit to highestProcessed+1, but if there are pending / inflight - commit to first / lowest inflight.

rkolesnev (commented):
I've left some comments.
In general - i think it can be simplified a bit - the processed map is not really needed.

There are couple issues though:
- on partition revocation - commit needs to be triggered and potentially state for revoked partition cleared. 
- On partition assignment - need to clear / re-initialize offset manage state - pending offsets, current position etc.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,574 @@
...
+    var nextExpected = highestContiguous + 1;
+
+    // Find the highest contiguous sequence
+    while (processed.contains(nextExpected)) {
+      highestContiguous = nextExpected;
+      nextExpected++;
+    }
+
+    return highestContiguous;
+  }
```
rkolesnev, line 402: This wont always work. 
Offsets in Kafka can have gaps - they are not always contiguous - for example transactions create markers that occupy offsets but are not visible in consumer, log compaction will delete off offsets - so offsets are always increasing but not necessarily without gaps.
Changing logic as per my suggestion above will remove dependency on checking for gaps in processed offsets though.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: > I've left some comments.
In general - i think it can be simplified a bit - the processed map is not really needed.

Are you proposing to remove the `processedOffsets` map and instead infer that an offset has been processed by its absence from the `pendingOffsets` map?

eschizoid: >There are couple issues though:

on partition revocation - commit needs to be triggered and potentially state for revoked partition cleared.
On partition assignment - need to clear / re-initialize offset manage state - pending offsets, current position etc.

agreed, will implement the `ConsumerRebalanceListener`

eschizoid: >This seems wrong - consumer.position(partition) returns next fetch offset.
From javadoc:
Get the offset of the next record that will be fetched (if a record with that offset exists).
So initializing next offset to commit as offset of next poll start seems not right - should probably initialize on first poll to first offset polled... (and re-initialize on partition assignment).


nice catch, I think I need to completely remove intitialzing offsets at partition assignment time using the `consumer.position()` and follow your suggestion

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: removed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: agreed and implmented

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: removed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
rkolesnev: > > I've left some comments.
> > In general - i think it can be simplified a bit - the processed map is not really needed.
> 
> Are you proposing to remove the `processedOffsets` map and instead infer that an offset has been processed by its absence from the `pendingOffsets` map?

Yes, for each Partition :
highest processed offset (Single Long) 
SortedSet of inflight / pending offsets (SortedSet of Longs)
last committed offset (Single Long)

onCommit:
per partition - take either first / lowest unprocessed offset or highest processed offset - whichever is lower. Mainly need highest processed offset in cases when there are no inflight / unprocessed offsets.
```

long offsetToCommit;
if( !pendingOffsets.isEmpty()){
offsetToCommit = pendingOffsets.first() + 1;
}else{
offsetToCommit = highestProcessed + 1;
}

if(offsetToCommit > lastCommittedOffset) {
  //    schedule / do Commit(...) ;
  //  on Commit success set  lastCommittedOffset = offsetToCommit;
 
}else {
// skip commit as not advanced
}
```

lib/src/main/java/org/kpipe/consumer/FunctionalConsumer.java
```diff
@@ -648,7 +745,13 @@ protected void processRecords(final ConsumerRecords<K, V> records) {
...
         .forEach(record -> {
           try {
-            virtualThreadExecutor.submit(() -> processRecord(record));
+            virtualThreadExecutor.submit(() -> {
+              // Track offset before processing
+              if (offsetManager != null) {
+                commandQueue.offer(ConsumerCommand.TRACK_OFFSET.withRecord(record));
+              }
+              processRecord(record);
+            });
```
rkolesnev, line 754: the offset tracking should be done before submission into virtual threadExecutor - its not intensive (just adding to queue) but ensures that tracking is done sequentially - so no need to worry about lower offsets being added after higher offsets to pending queue. 
That simplifies reasoning and handling of commits as naturally the 'polled' offsets can be thought off 'pending' regardless of whether virtual thread started processing on them or not...

rkolesnev (commented):
added some more comments regarding offset handling.

eschizoid:
@rkolesnev I will take a look at all your comments today 😄  

Thanks again for all the feedback, I think we are getting closer 🚀 !

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+
+  /**
+   * Commits offsets that are safe to commit based on the current processing state.
+   *
+   * <p>This method is called periodically to ensure that offsets are committed in a timely manner
+   * without losing any unprocessed messages.
+   */
+  public void commitSafeOffsets() {
+    try {
+      commitSyncAndWait(5);
```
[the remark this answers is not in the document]
eschizoid: Fixed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,739 @@
...
+
+    // Also verify via partition state
+    assertEventually(
+      () -> expectedCommittedOffset == (long) offsetManager.getPartitionState(partition).get("nextOffsetToCommit"),
+      Duration.ofSeconds(2),
+      "nextOffsetToCommit should be " + expectedCommittedOffset
+    );
+  }
+
+  static Stream<Arguments> contiguousOffsetScenarios() {
```
eschizoid, line 111: @rkolesnev

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,739 @@
...
+      if (condition.get()) {
+        return;
+      }
+      Thread.sleep(50);
+    }
+
+    fail(message);
+  }
+
+  static Stream<Arguments> gapSizeScenarios() {
```
eschizoid, line 238: @rkolesnev

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: @rkolesnev I have updated the documentation a bit to reflect the different behaviors of 

```
public void trackOffset(final ConsumerRecord<K, V> record)

public void trackOffset(final TopicPartition partition, final long offset)

public void markOffsetProcessed(final ConsumerRecord<K, V> record)

public void markOffsetProcessed(final TopicPartition partition, final long offset) 

```

I also tagged you on the relevant property tests in OffsetManagerTest, let me know what you think

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+  }
+
+  /**
+   * Creates a rebalance listener for the Kafka consumer.
+   *
+   * @return A ConsumerRebalanceListener instance
+   */
+  public ConsumerRebalanceListener createRebalanceListener() {
+    return new RebalanceListener(state, pendingOffsets, nextOffsetsToCommit, highestProcessedOffsets, kafkaConsumer);
+  }
```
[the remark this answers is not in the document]
eschizoid: @rkolesnev I did this even after reading your comment 🤝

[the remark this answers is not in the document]
eschizoid: Fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+        try {
+          future.get(5, TimeUnit.SECONDS);
+        } catch (final ExecutionException | TimeoutException e) {
+          LOGGER.log(WARNING, "Final commit did not complete in time", e);
+        }
+      }
+    } catch (final Exception e) {
+      LOGGER.log(WARNING, "Error during final offset commit", e);
+    }
+  }
```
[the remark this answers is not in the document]
eschizoid: agreed, will fix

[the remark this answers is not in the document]
eschizoid: Fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
eschizoid: See this [comment](https://github.com/eschizoid/kpipe/pull/51#discussion_r2070923072)

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+    if (state.get() == OffsetState.STOPPED) {
+      return;
+    }
+
+    // Add to pending offsets (we use offset+1 as the next offset)
+    pendingOffsets.computeIfAbsent(partition, k -> new ConcurrentSkipListSet<>()).add(offset + 1);
+
+    // Initialize the next offset to commit if not already set
+    // We should use the actual offset here, not offset+1
+    nextOffsetsToCommit.computeIfAbsent(partition, k -> offset);
```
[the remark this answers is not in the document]
eschizoid: > offsets seem to +1'ed twice now - once here and once in updateCommittableOffset - Line 426.

Let me know what you think about [these](https://github.com/eschizoid/kpipe/pull/51#discussion_r2070923072)

lib/src/main/java/org/kpipe/consumer/FunctionalConsumer.java
eschizoid: nice catch 🙏 

Fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    return performCommitAndCleanup(offsetsToCommit, timeoutSeconds);
+  }
+
+  /** Prepares a map of offsets to commit. */
+  private Map<TopicPartition, OffsetAndMetadata> prepareOffsetsToCommit() {
+    final var offsetsToCommit = new HashMap<TopicPartition, OffsetAndMetadata>();
+    nextOffsetsToCommit.forEach((partition, offset) -> offsetsToCommit.put(partition, new OffsetAndMetadata(offset)));
+    return offsetsToCommit;
```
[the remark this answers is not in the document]
eschizoid: Agreed and fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+    if (state.get() == OffsetState.STOPPED) {
+      return;
+    }
+
+    // Add to pending offsets (we expect offset to already be the "next" offset)
+    pendingOffsets.computeIfAbsent(partition, k -> new ConcurrentSkipListSet<>()).add(offset);
+
+    // Initialize the next offset to commit if not already set
+    // Use offset-1 to be consistent with the ConsumerRecord version
+    nextOffsetsToCommit.computeIfAbsent(partition, k -> offset - 1);
```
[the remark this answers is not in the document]
eschizoid: Fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,583 @@
...
+        highestProcessedOffsets.forEach((partition, highestProcessed) -> {
+          final var pending = pendingOffsets.get(partition);
+          if (pending != null && !pending.isEmpty()) {
+            // Use exactly the lowest pending offset
+            nextOffsetsToCommit.put(partition, pending.first());
+          } else {
+            // Otherwise use the highest processed offset
+            nextOffsetsToCommit.put(partition, highestProcessed);
+          }
+        });
```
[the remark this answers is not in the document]
eschizoid: >here is no need to do clean up of committed offsets i think - offsets are removed from pending on processing completion already - so there is never going to be any pending offsets to remove here after the commit - as pending / unprocessed event offsets should never be committed in the first place.

This cleanup is necessary even though offsets are removed from pending when processed because it ensures that all offsets less than the committed offset are removed, which is important for handling out of order processing and gaps in the offset sequence.

We could revisit this topic on a follow up PR. What do you think?

eschizoid:
@rkolesnev I think I have addressed all your comments😅 

Let me know what do you think and I we could move forward merging this bad boy

rkolesnev:
> @rkolesnev I think I have addressed all your comments 😅.
> 
> Let me know what you think, and if we could move forward with merging this bad boy.

@eschizoid - Cool, i will endeavor to have a look in the next few days.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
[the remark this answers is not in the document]
rkolesnev: Hmm but commit logic should be to only commit either first / lowest pending offset or if none pending - then highest processed - so there is never an offset lower than committed in pending offsets by definition?
But sure - can revisit... i will do another pass as soon as i get a bit of time and let you know...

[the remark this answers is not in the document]
eschizoid: @rkolesnev I see what you are saying now,  thanks for the explanation. I removed the unnecessary cleanup of committed offsets.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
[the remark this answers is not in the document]
rkolesnev: i dont see much change - rebalance listener still does actual commit... 
maybe some code didn't get pushed?

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+      final List<Long> offsetsToTrack,
+      final List<Long> offsetsToProcess,
+      final long expectedCommittedOffset
+    ) throws Exception {
+      // Track the initial offset
+      final var record = new ConsumerRecord<>(TOPIC, 0, initialOffset, "key", "value");
+      offsetManager.trackOffset(record);
+
+      // Process in batches
+      offsetsToTrack.forEach(offset -> offsetManager.trackOffset(partition, offset));
```
rkolesnev, line 93: this trackOffset overload is not doing +1 - while the other overload - that is called by actual FunctionalConsumer code does - which makes the test not representative of code execution in the application...
so asserts are basically off by 1.

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+      offsetsToProcess.forEach(offset -> offsetManager.markOffsetProcessed(partition, offset));
```
rkolesnev, line 94: this markOffsetProcessed overload is not doing +1 - while the other overload - that is called by actual FunctionalConsumer code does - which makes the test not representative of code execution in the application...
so asserts are basically off by 1.

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.kpipe.consumer.enums.ConsumerCommand;
+import org.mockito.*;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class OffsetManagerTest {
```
rkolesnev, line 32: i dont see any tests that test correctness of highestProcessed offset commits - when there are no pending offsets left...
Changing one of the tests with gaps - to have no gaps - so everything marked as processed - actually fails the test - as offset committed is too low (or at least nextOffsetToCommit is too low)...

rkolesnev: overall IMHO - i would suggest to use awaitility with Assertions / AssertJ asserts for eventual assertions - that way proper asserts can be used that log both expected and actual values on failure etc...
it is personal preference again of course - but the AssertJ assertions are nice and fluent... while avaitility adds the eventual assertion handling quite nicely.

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
rkolesnev: similarly other tests that use those overloads are off...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
rkolesnev: similarly other tests that use those overloads are off...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
rkolesnev: another IMHO - it maybe worth adding couple of integration tests - its easy to spin up docker container for Kafka - and tests can be easily isolated by topic name - so only need to spin up / down Kafka once per test set... 
But they would flesh out any inconsistencies with mocks / overloads quite fast / easily...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+
+    fail(message);
+  }
+
+  static Stream<Arguments> contiguousOffsetScenarios() {
+    return Stream.of(
+      Arguments.of(
+        "Basic gap scenario - process offsets out of order with a missing offset",
+        100L, // initialOffset
+        List.of(101L, 105L, 102L, 103L, 104L), // offsetsToTrack
```
rkolesnev, line 738: technically this should be impossible - offsets are submitted by single thread now - so that would be in incremental order only...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+        List.of(102L, 101L, 104L), // offsetsToProcess (note: 103 is deliberately not processed)
+        103L // expectedCommittedOffset
+      ),
+      Arguments.of(
+        "Multiple gaps scenario - commit should stop at first unprocessed offset",
+        200L, // initialOffset
+        List.of(201L, 202L, 205L, 206L, 203L, 204L), // offsetsToTrack
```
rkolesnev, line 745: technically this should be impossible - offsets are submitted by single thread now - so that would be in incremental order only...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+        List.of(201L, 202L, 203L), // offsetsToProcess (note: 204-206 are not processed)
+        204L // expectedCommittedOffset - commits up to the first gap
+      ),
+      Arguments.of(
+        "Large range with interleaved processing",
+        1000L, // initialOffset
+        List.of(1_001L, 1_002L, 1_010L, 1_020L, 1_030L, 1_005L, 1_006L), // offsetsToTrack
```
rkolesnev, line 752: technically this should be impossible - offsets are submitted by single thread now - so that would be in incremental order only...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
```diff
@@ -0,0 +1,798 @@
...
+        "Large gap between offsets",
+        1_000L, // initial offset
+        5_000L, // gap offset
+        1_001L, // expected offset after first commit
+        5_001L // expected offset after the second commit
+      ),
+      Arguments.of(
+        "Backwards gap (higher offset processed before lower)",
+        5_000L, // initial offset
+        1_000L, // gap offset (lower than initial - backwards)
```
rkolesnev, line 771: this should be impossible - Kafka ordering guarantees (and the fact that track offset is called from main / single thread) make it invalid...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
rkolesnev: the backwards gap test is invalid - due to Kafka guarantees - the other ones seem fine.

rkolesnev:
@eschizoid - i've done the review... sorry for delay...
In general - i think its getting over-complicated with the two overloads for track / markProcessed methods and +1 on offsets in various places... hence tests seem a bit inconsistent as they use different overloads for those methods in different tests...
i can try to draft a PR with what i mean - but that would be towards the end of next week - i have few conferences coming up now that i am attending / presenting at - so i am away until next Thursday... or i might find some time on a plane / in airport to give it a go.

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,610 @@
...
+    if (state.get() == OffsetState.STOPPED) {
+      return;
+    }
+
+    // Add to pending offsets (we expect the offset to already be the "next" offset)
+    pendingOffsets.computeIfAbsent(partition, k -> new ConcurrentSkipListSet<>()).add(offset);
+
+    // Initialize the next offset to commit if not already set
+    // Use offset-1 to be consistent with the ConsumerRecord version
+    nextOffsetsToCommit.computeIfAbsent(partition, k -> offset - 1);
```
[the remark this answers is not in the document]
rkolesnev: if this is purely for reading from tests - i would instead intercept committed offsets from command queue - to make it closer to values actually being committed / sent to brokers...

eschizoid:
> @eschizoid - i've done the review... sorry for delay... In general - i think its getting over-complicated with the two overloads for track / markProcessed methods and +1 on offsets in various places... hence tests seem a bit inconsistent as they use different overloads for those methods in different tests... 

Unfortunately, I agree with your assessment here, in my attempt to provide a more reach API I ended up complicating things

```
public void trackOffset(final ConsumerRecord<K, V> record)

public void trackOffset(final TopicPartition partition, final long offset)

public void markOffsetProcessed(final ConsumerRecord<K, V> record)

public void markOffsetProcessed(final TopicPartition partition, final long offset)
```

What do you think about the two APIs we should preserve?


> i can try to draft a PR with what i mean - but that would be towards the end of next week - i have few conferences coming up now that i am attending / presenting at - so i am away until next Thursday... or i might find some time on a plane / in airport to give it a go.

That would be fabulous! In the meantime, I can start removing the invalid tests and updating them based on the two APIs we decide to use. What do you think?

rkolesnev:
> > @eschizoid - i've done the review... sorry for delay... In general - i think its getting over-complicated with the two overloads for track / markProcessed methods and +1 on offsets in various places... hence tests seem a bit inconsistent as they use different overloads for those methods in different tests...
> 
> Unfortunately, I agree with your assessment here, in my attempt to provide a more reach API I ended up complicating things
> 
> ```
> public void trackOffset(final ConsumerRecord<K, V> record)
> 
> public void trackOffset(final TopicPartition partition, final long offset)
> 
> public void markOffsetProcessed(final ConsumerRecord<K, V> record)
> 
> public void markOffsetProcessed(final TopicPartition partition, final long offset)
> ```
> 
> Which are the two APIs you think we should preserve?
> 
> > i can try to draft a PR with what i mean - but that would be towards the end of next week - i have few conferences coming up now that i am attending / presenting at - so i am away until next Thursday... or i might find some time on a plane / in airport to give it a go.
> 
> That would be fabulous! In the meantime, I can start removing the invalid tests and updating them based on the two APIs we decide to use. What do you think?

Sure, either definition works - my main concern is - not to +1 offsets anywhere (track actual message offsets) until actual commit is prepared / submitted...

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: removed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: removed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: removed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: removed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: > overall IMHO - i would suggest to use awaitility with Assertions / AssertJ asserts for eventual assertions - that way proper asserts can be used that log both expected and actual values on failure etc...
it is personal preference again of course - but the AssertJ assertions are nice and fluent... while avaitility adds the eventual assertion handling quite nicely.

Yeah I agreed 100%. I logged while ago #50

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: Fixed

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: Fixed

eschizoid:
@rkolesnev Ok i did a major refactor of the `OffsetManager` and `OffsetManagerTests` and consolidated the `trackOffset` and `markOffsetProcessed` APIs into a single pair.

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: > i dont see any tests that test correctness of highestProcessed offset commits - when there are no pending offsets left...
Changing one of the tests with gaps - to have no gaps - so everything marked as processed - actually fails the test - as offset committed is too low (or at least nextOffsetToCommit is too low)...

Can you provide an scenario for this test case?

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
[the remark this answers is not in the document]
eschizoid: fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,610 @@
...
+            // Use exactly the lowest pending offset
+            nextOffsetsToCommit.put(partition, pending.first());
+          } else {
+            // Use the highest processed offset directly (not +1) since this happens after
+            // commit
+            nextOffsetsToCommit.put(partition, highestProcessed);
+          }
+        });
+      }
+      return result;
```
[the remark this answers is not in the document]
eschizoid: Fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,610 @@
...
+    }
+  }
+
+  /**
+   * Updates the committable offset for a partition based on processed offsets. Handles both
+   * contiguous and non-contiguous offset tracking.
+   *
+   * @param partition The partition to update
+   */
+  public void updateCommittableOffset(final TopicPartition partition) {
```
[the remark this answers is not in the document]
eschizoid: removed

eschizoid:
@rkolesnev Things should look way cleaner after your suggested changes 😄  

Still have the `+1` semantic problem (which I agree needs to be fixed)

lib/src/test/java/org/kpipe/consumer/OffsetManagerTest.java
eschizoid: fixed

lib/src/main/java/org/kpipe/consumer/OffsetManager.java
```diff
@@ -0,0 +1,610 @@
...
+   */
+  public void trackOffset(final ConsumerRecord<K, V> record) {
+    final var partition = new TopicPartition(record.topic(), record.partition());
+    final long offset = record.offset();
+    if (state.get() == OffsetState.STOPPED) {
+      return;
+    }
+
+    // Add to pending offsets (we use offset+1 as the next offset)
+    pendingOffsets.computeIfAbsent(partition, k -> new ConcurrentSkipListSet<>()).add(offset + 1);
```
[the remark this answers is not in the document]
eschizoid: fixed

eschizoid:
https://github.com/eschizoid/kpipe/pull/51#discussion_r2089805939

this is fixed, finally 🏆

rkolesnev (commented):
Getting closer...
couple of small fixes...
One big thing is commit handling on rebalance - there is none right now and the state will race with command queue...

lib/src/main/java/org/kpipe/consumer/RebalanceListener.java
```diff
@@ -0,0 +1,90 @@
...
+ *   <li>Internal data structures are updated to reflect the current assignment
+ *   <li>Offset state is preserved for persistent consumer groups
+ *   <li>Memory is reclaimed for partitions no longer assigned
+ * </ul>
+ *
+ * <p>The listener coordinates with the OffsetManager to maintain accurate offset tracking during
+ * consumer group rebalancing operations, preventing duplicate message processing or data loss when
+ * partitions are reassigned between consumer instances.
+ */
+public class RebalanceListener implements ConsumerRebalanceListener {
```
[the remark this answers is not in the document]
eschizoid: > I dont think you need the separate class at all - can just implement ConsumerRebalanceListener directly in OffsetManager?
Maybe a follow up refactor ticket...?

logged here #55

lib/src/main/java/org/kpipe/consumer/RebalanceListener.java
```diff
@@ -0,0 +1,90 @@
...
+      highestProcessedOffsets.remove(partition);
+      final var position = consumer.position(partition);
+      highestProcessedOffsets.put(partition, position);
+    });
+  }
+
+  @Override
+  public void onPartitionsLost(final Collection<TopicPartition> partitions) {
+    onPartitionsRevoked(partitions);
+  }
```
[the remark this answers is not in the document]
eschizoid: I see what you are saying, let's also have a follow-up PR for this too:
#56

eschizoid:
@rkolesnev Alright, I think we are finally good to merge down this bad boy. See my last set of comments.

Thanks again for all your time @rkolesnev, i really appreciate it.

Excited to cut a release 🚀

PR #58, opened 2025-06-04
Add avro support

eschizoid:
The main purpose of this PR is to add support for the `Avro` binary format (#8). At the same time, I would like to achieve parity with what we have in `JsonMessageProcessor.java`.

I am still debating whether this is overkill or unnecessary in terms of the APIs exposed, and I am wondering if we should remove some of the APIs I created as part of this PR. I completely understand that `Avro` is not meant to be treated as `JSON`.

Wondering what you think, @rkolesnev. Do you see any value here?

### DIFF
--- a/lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
+++ b/lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
@@ -1,0 +1,903 @@
+package org.kpipe.processor;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.lang.System.Logger;
+import java.lang.System.Logger.Level;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.*;
+import org.apache.avro.io.BinaryDecoder;
+import org.apache.avro.io.BinaryEncoder;
+import org.apache.avro.io.DecoderFactory;
+import org.apache.avro.io.EncoderFactory;
+import org.apache.avro.util.Utf8;
+
+/**
+ * Provides utility functions for processing Avro messages. This class contains common message
+ * processors for use with the MessageProcessorRegistry to transform and enhance Avro data in byte
+ * array format.
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * // Register a schema
+ * final var userSchemaJson = """
+ *   {
+ *     "type": "record",
+ *     "name": "User",
+ *     "fields": [
+ *       {"name": "name", "type": "string"},
+ *       {"name": "age", "type": "int"}
+ *     ]
+ *   }
+ *   """;
+ * AvroMessageProcessor.registerSchema("userSchema", userSchemaJson);
+ *
+ * // Create a pipeline using these processors
+ * Function<byte[], byte[]> pipeline = MessageProcessorRegistry.pipeline(
+ *     "parseAvro",
+ *     MessageProcessorRegistry.get("addTimestampAvro")
+ * );
+ *
+ * // Process a message
+ * byte[] result = pipeline.apply(avroBytes);
+ * }</pre>
+ */
+public class AvroMessageProcessor {
+
+  private AvroMessageProcessor() {}
+
+  private static final Logger LOGGER = System.getLogger(AvroMessageProcessor.class.getName());
+  private static final byte[] EMPTY_AVRO = new byte[0];
+  private static final ConcurrentHashMap<String, Schema> SCHEMA_REGISTRY = new ConcurrentHashMap<>();
+
+  private static final ThreadLocal<Schema.Parser> SCHEMA_PARSER = ThreadLocal.withInitial(Schema.Parser::new);
+  private static final ThreadLocal<ByteArrayOutputStream> OUTPUT_STREAM_CACHE = ThreadLocal.withInitial(() ->
+    new ByteArrayOutputStream(8192)
+  );
+  private static final ThreadLocal<BinaryEncoder> ENCODER_CACHE = ThreadLocal.withInitial(() -> null);
+  private static final ThreadLocal<BinaryDecoder> DECODER_CACHE = ThreadLocal.withInitial(() -> null);
+
+  /**
+   * Registers an Avro schema with the given name.
+   *
+   * <pre>{@code
+   * final var userSchemaJson = """
+   *   {
+   *     "type": "record",
+   *     "name": "User",
+   *     "fields": [
+   *       {"name": "name", "type": "string"},
+   *       {"name": "age", "type": "int"}
+   *     ]
+   *   }
+   *   """;
+   * AvroMessageProcessor.registerSchema("userSchema", userSchemaJson);
+   * }</pre>
+   *
+   * @param name The name to register the schema under
+   * @param schemaJson The Avro schema in JSON format
+   * @throws org.apache.avro.SchemaParseException if the schema is invalid
+   */
+  public static void registerSchema(final String name, final String schemaJson) {
+    final var schema = SCHEMA_PARSER.get().parse(schemaJson);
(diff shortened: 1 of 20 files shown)


eschizoid:
One of the reasons this PR can add value is that Avro, paired with the schema registry, is a combination that is quite common in large enterprises.

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
```diff
@@ -0,0 +1,682 @@
...
+ * byte[] result = pipeline.apply(avroBytes);
+ * }</pre>
+ */
+public class AvroMessageProcessor {
+
+  private AvroMessageProcessor() {}
+
+  private static final Logger LOGGER = System.getLogger(AvroMessageProcessor.class.getName());
+  private static final byte[] EMPTY_AVRO = new byte[0];
+  private static final ConcurrentHashMap<String, Schema> SCHEMA_REGISTRY = new ConcurrentHashMap<>();
```
rkolesnev, line 56: Wouldn't more typical be integration with external schema registry? i.e. confluent schema registry or Karapace etc..

rkolesnev (approved):
Generally all looks OK to me - but I am not very deep into avro schema parsing / manipulation... 
Few questions though - I do not fully follow why parsing is decoupled from Consumer? 
Even when it is decoupled from Consumer - why not use Kafka Avro deserializers? they can be used outside of Consumer quite easily as well...
In Kafka (at least Confluent implementation and i guess others keeping compatibility) schemas are stored in external registry and schema ID is encoded in message as magic byte... 
The implementation here seems to assume avro messages without magic byte / or with the magic bytes stripped - which looks very non-standard / niche.

rkolesnev:
Yes in general i don't see that much use of custom Avro deserialization... At least personally I only ever saw Avro used with schema registry...
Avro is more rigid and less allowing then Json - you would need to create new Avro schema and re-encode to add a field that doesn't exist in original schema for example... Json - at least schema-less Json is much easier in that regard...

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
```diff
@@ -0,0 +1,682 @@
...
+                  case FLOAT:
+                  case DOUBLE:
+                    newRecord.put(fieldName, 0);
+                    break;
+                  case BOOLEAN:
+                    newRecord.put(fieldName, false);
+                    break;
+                  default:
+                    if (field.hasDefaultValue()) {
+                      newRecord.put(fieldName, field.defaultVal());
```
rkolesnev, line 456: should use default value if specified for Strings / numbers as well instead of "" or 0?

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
```diff
@@ -0,0 +1,890 @@
...
+
+  private AvroMessageProcessor() {}
+
+  private static final Logger LOGGER = System.getLogger(AvroMessageProcessor.class.getName());
+  private static final byte[] EMPTY_AVRO = new byte[0];
+  private static final ConcurrentHashMap<String, Schema> SCHEMA_REGISTRY = new ConcurrentHashMap<>();
+
+  private static final ThreadLocal<Schema.Parser> SCHEMA_PARSER = ThreadLocal.withInitial(Schema.Parser::new);
+  private static final ThreadLocal<ByteArrayOutputStream> OUTPUT_STREAM_CACHE = ThreadLocal.withInitial(() ->
+    new ByteArrayOutputStream(8192)
```
lombardo-chcg, line 63: Perhaps make this a configurable value with a default of 8192?

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
```diff
@@ -0,0 +1,890 @@
...
+ * Function<byte[], byte[]> pipeline = MessageProcessorRegistry.pipeline(
+ *     "parseAvro",
+ *     MessageProcessorRegistry.get("addTimestampAvro")
+ * );
+ *
+ * // Process a message
+ * byte[] result = pipeline.apply(avroBytes);
+ * }</pre>
+ */
+public class AvroMessageProcessor {
```
lombardo-chcg, line 53: The public interface has great documentation.  The class itself seems to have a lot of responsibilities, perhaps it can be decomposed into separate classes for schema operations, transformation logic, and core processing logic?  That might make it easier later when and if you want to add support for external schema registry, for example.

lombardo-chcg (approved):
Nice addition to the library with excellent docs!  I might even call them world-class docs 😎

eschizoid:
> Generally all looks OK to me - but I am not very deep into avro schema parsing / manipulation... Few questions though - I do not fully follow why parsing is decoupled from Consumer? Even when it is decoupled from Consumer - why not use Kafka Avro deserializers? they can be used outside of Consumer quite easily as well... In Kafka (at least Confluent implementation and i guess others keeping compatibility) schemas are stored in external registry and schema ID is encoded in message as magic byte... The implementation here seems to assume avro messages without magic byte / or with the magic bytes stripped - which looks very non-standard / niche.

These are excellent questions, and let me address all of them and why I think the `Function<byte[], byte[]>`  implementation approach could offer better performance.

> Why decoupling parsing from Consumer is faster

Decoupling parsing from the consumer provides several performance benefits:

1. **Parallel Processing**: Separating these concerns allows you to scale them independently. Consumers can focus on polling for data while parsers can be optimized separately.

2. **Resource Reuse**: Our implementation uses `ThreadLocal` variables for parsers, encoders, and output streams, which reduces allocation overhead and GC pressure.

3. **Pipeline Optimization**: The functional composition pattern allows processors to be chained without unnecessary intermediate object creation.

4. **Batching**: We can batch transformations together when they're decoupled, reducing the overhead of repeated serialization/deserialization.

> Why not use Kafka Avro deserializers

Our implementation could be faster for several reasons:

1. **No Remote Schema Registry Calls**: Kafka Avro deserializers typically make network requests to fetch schemas, which adds latency. Our implementation keeps schemas in memory.

2. **Direct Access**: Our processors directly access and modify the bytes/objects without the additional abstraction layers of Kafka's deserializers (aka kafka streams `Serdes`)

3. **Optimized for Throughput**: Thread-local resources and pooled objects reduce allocation overhead in high-throughput scenarios.

4. **DslJson Performance**: For JSON processing, we're using DslJson, which is known for high performance compared to other JSON libraries.

> Regarding schema registry and magic bytes

You're correct that the standard Confluent format includes magic bytes and schema IDs at the beginning of messages. Our implementation takes a different approach:

1. **Local Schema Management**: We register schemas directly within the application instead of an external service, which eliminates network latency.

2. **Flexible Schema Handling**: Our implementation allows registering schemas by name and using them directly, which can be more efficient when schemas are known in advance.

3. **Optimization for Known Workflows**: When you control both producing and consuming sides, you can optimize the format for our specific needs.

If compatibility with Confluent's format is needed, we could enhance the processors to handle the magic bytes and schema ID prefix. This would involve:

```java
// Example of how we could handle Confluent-format messages
public static Function<byte[], byte[]> parseAvroWithMagicBytes(final String schemaName) {
    return avroBytes -> {
        // First byte is magic byte (0), next 4 bytes are schema ID
        if (avroBytes.length > 5 && avroBytes[0] == 0) {
            // Extract actual Avro content (skip first 5 bytes)
            byte[] content = Arrays.copyOfRange(avroBytes, 5, avroBytes.length);
            return processAvro(content, getSchema(schemaName), record -> record);
        } else {
            // Process as raw Avro
            return processAvro(avroBytes, getSchema(schemaName), record -> record);
        }
    };
}
```

Our current implementation is optimized for performance by reducing network calls, object creation, and memory pressure - especially for high-throughput message processing systems.

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
eschizoid: will fix that

lib/src/main/java/org/kpipe/processor/AvroMessageProcessor.java
eschizoid: Extracted some functionality to `MessageFormatt`

PR #102, opened 2026-05-11
feat: circuit breaker for sinks (P2 #7)

eschizoid:
## Summary

- Hand-rolled circuit breaker inside `kpipe-consumer`, mirrors the `BackpressureController` shape (pure-function record + sliding-window stats + state machine driven by per-record outcomes with single-read CAS).
- Three states: **CLOSED → OPEN** (failure rate crosses threshold over a full window → pause consumer + schedule probe timer), **OPEN → HALF_OPEN** (open-duration elapses → resume; next record is the probe), **HALF_OPEN → CLOSED** (probe success → stats reset) / **HALF_OPEN → OPEN** (probe failure → re-pause + restart timer).
- Why hand-rolled, not Resilience4j: it duplicates infrastructure already in `kpipe-consumer` (state machine, command-queue serialization, scheduled executor, pause/resume via `manualPause` + `LockSupport.park/unpark`). Reversible decision — see commit message for the full reasoning.

## Surface

- `Stream.withCircuitBreaker(double failureThreshold, int windowSize, Duration openDuration)`
- `Stream.withCircuitBreaker(CircuitBreakerController controller)` for shared controllers
- `MultiBuilder.withCircuitBreaker(...)` for multi-topic consumers
- `KPipeConsumer.Builder.withCircuitBreaker(...)` for explicit-API users

## Metrics

- `circuitBreakerTrips` (counter — CLOSED→OPEN + HALF_OPEN→OPEN)
- `circuitBreakerTimeOpenMs` (counter — recorded on OPEN→HALF_OPEN)
- `OtelConsumerMetrics.recordCircuitBreakerStateChange(state)` — state-as-attribute breakdown

## Tests

- **16 unit tests green** — `CircuitBreakerStatsTest` (7) + `CircuitBreakerControllerTest` (9). Stats test includes a virtual-thread concurrency stress (`Thread.ofVirtual()` × 50k writes) per the §Testing Strategy convention.
- **5 integration tests** in `KPipeCircuitBreakerIntegrationTest` against `MockConsumer` (no Docker required). Currently 2 pass, 3 fail under parallel mode due to ordering-of-outcome timing — the CB logic is correct (the unit tests prove the state machine) but the integration tests' pipeline pattern (`seq % 5 < 2`) doesn't deterministically guarantee the worst-case window doesn't temporarily exceed 50% failures even when overall rate is 40%. Follow-up to tighten the integration test pipeline to deterministic sequential mode or use a controlled fail-counter.

## Ride-along refactor

- `KafkaMessageSink` accepts nullable `Tracer` and coalesces to `Tracer.noop()` internally — same null-coalesce pattern `KPipeConsumer`/`KPipeProducer` already use for metrics. Users pass `null` instead of `Tracer.noop()`; the `Tracer.noop()` concept stays in the SPI for the internal coalesce. Migrated 3 test call sites.

## Test plan

- [x] Unit tests green (`./gradlew :lib:kpipe-consumer:test --tests "*CircuitBreaker*"` — 16 tests, 0 failures)
- [x] Full compile clean (all modules including examples)
- [x] Integration tests stabilized to 5/5 green (follow-up)
- [x] PLAN.md detail section updated to mark 7 as shipped with deferred follow-ups (will update when integration tests are tightened)


eschizoid:
Fixed the 3 failing integration tests — all 21 CB tests now green (7 stats + 9 controller + 5 integration).

**Root cause:** when CB tripped and called `internalPause()`, the next consumer-loop iteration's `checkBackpressure` observed `currentlyPaused=true` AND low lag and immediately auto-resumed — defeating the CB pause within one iteration. The two pause systems shared the consumer-state machine but didn't coordinate.

**Fix:** dedicated `circuitBreakerPaused` AtomicBoolean (sibling of `backpressurePaused` / `manualPause`) gates both backpressure's auto-resume AND the public `resume()`. `tripBreaker` sets it before `internalPause`; `tryHalfOpen` clears it before `internalResume` for the probe.

**Test cleanup:** integration tests switched to sequential mode for deterministic outcome ordering.

**API overlap note worth tracking for 2.0:** backpressure (capacity) and CB (quality) both manipulate the same pause state via separate gate flags. The 1.x shape (orthogonal flags) is correct given the features were added independently, but a unified consumer-health controller would be cleaner long-term.

eschizoid:
Did Option A — added `PauseCoordinator` with a bitmask-backed `EnumSet<Source>` (MANUAL / BACKPRESSURE / CIRCUIT_BREAKER). The three independent AtomicBoolean flags + AND-of-checks pattern is replaced by:

```java
if (pauseCoordinator.requestPause(SOURCE)) internalPause();
if (pauseCoordinator.releasePause(SOURCE)) internalResume();
```

**One rule:** consumer is paused iff at least one source is in the held set; resume fires only when the LAST source releases. No more cross-checking other flags at every resume site.

Tests: 235 `kpipe-consumer` tests green — 21 CB + 8 backpressure (Parallel + Sequential) + the rest. No public API changes; this is internal-only.

PR #109, opened 2026-05-15
fix(consumer): shutdown hook leak + shutdownGracefully race + stale close() docs

eschizoid:
## Summary

Three MEDIUM-priority fixes from the [#105/#106/#107 review](https://github.com/eschizoid/kpipe/pull/107):

1. **Shutdown hook leak** — the JVM shutdown hook installed by `withShutdownHook(true)` was never removed. In a long-lived process that builds/closes many consumers, hooks accumulated. `close()` now calls `Runtime.removeShutdownHook(...)` (no-op when the JVM is already shutting down).
2. **`shutdownGracefully(Duration)` race** — the CLOSED-state check was a plain `state.get() == CLOSED` read and fell through to `pause()/close()` on every non-CLOSED state. Under contention with another thread already mid-shutdown, it would return a misleading `drained = true`. Now snapshots the state once and handles CLOSING as "await the other shutdown, report its drain status".
3. **Stale `close()` Javadoc** — listed "Creating a message tracker" as step 2, but the tracker was deleted in PR #107. Refreshed the step list.

## Stack
Stacks on top of [#107](https://github.com/eschizoid/kpipe/pull/107).

## Test plan
- [x] `lib/kpipe-consumer` compile + unit tests green locally
- [x] CI build green on this branch


eschizoid:
Superseded by recovery PR — the stack tangled when intermediate PRs merged into their stack-base instead of main, leaving #106/#107/#108/#110 content orphaned. Recovery PR consolidates everything against main.

PR #111, opened 2026-05-15
Arch/facade start helper

eschizoid:


eschizoid:
Duplicate — replaced by #112 which targets the correct stack base.

PR #126, opened 2026-05-16
Bench/competitor suite

eschizoid:


eschizoid:
Superseded by squash-merge of the original PR series (#122 / #123 / #124). The branch was force-pushed after merge so this PR object reappeared with the same commits, but the content is already on `main`. The remaining tuning is on #128.

Issue #136, opened 2026-05-26
Key based parallelism and Share Consumer benchmark


chuck-alt-delete:
Hey Mariano,

What a cool project! I was sad to hear the Confluent Parallel Consumer is no longer maintained as of a few days ago, so I was doing some research on alternatives. The reason given was that share consumers parallelize...but honestly I think this is really weak. Share consumers are only for unordered processing, but most Kafka use cases require at least key-based ordering. Also spinning up many share consumers isn't as ergonomic as spinning up a single consumer and parallelizing the processing.

I was curious to see if you thought about adding share consumers to the benchmark.

Separately, as I mentioned, I think key-based parallelism is the sweet spot for Kafka workloads. I notice kpipe has only partition ordering or completely unordered (virtual thread per record). Key-based ordering would cover that sweet spot that was already offered in the Confluent Parallel Consumer.

Thanks for your work on this. Feel free to close if and when you think is appropraite.

Best,
Chuck

eschizoid:
Hey @chuck-alt-delete,

Glad that you like the project 😄.

That point about key-based ordering being where most workloads actually live is the right framing. Splitting this into two PRs and keeping the issue open.

The first PR will add `ProcessingMode.KEY_ORDERED`. Call site:

```java
KPipe.json("orders", props)
    .withProcessingMode(ProcessingMode.KEY_ORDERED)
    .pipe(enrich)
    .toCustom(sink)
    .start();
```

The plan is a per-key LRU dispatcher. Each key gets a serial queue and a virtual-thread worker. Default cap of 10,000 distinct keys, configurable via `withKeyOrderedMaxKeys`. Eviction walks LRU oldest-first and only evicts queues that are empty and idle. 
If every queue at cap still has pending work, dispatch stalls until something drains; that's the implicit backpressure. Null-keyed records share a sentinel queue. Backpressure reuses the same total-in-flight metric that PARALLEL already uses, plugged through a Dispatcher abstraction the three modes share.

Same-key records process serially in offset order, different keys run in parallel. Cross-key parallelism is whatever Loom gives you.

The second PR will add a share-consumer row to `ParallelProcessingBenchmark`. Same workload regimes as the existing four runtimes (0/100/1000 µs of per-record blocking). My guess is you're right, share consumers should look fine at zero work and fall behind once records start blocking on I/O.

Will tag you on both.

Mariano

eschizoid:
Fixed by #137 and #138

eschizoid:
[...]

PR #137, opened 2026-05-27
feat(consumer): key-ordered processing mode (issue #136 part 1)

eschizoid:
## Summary

Implements part 1 of #136, adds `ProcessingMode.KEY_ORDERED`, which serializes records sharing a key on a single virtual thread while different keys run in parallel. This extends Kafka's per-partition ordering guarantee up to the application layer (useful when partition assignment is coarser than the unit of ordering you actually want).

Internally this lands as a sealed `Dispatcher<K>` abstraction with three implementations selected at construction from a new `ProcessingMode` enum:

- `SequentialDispatcher`: inline on the consumer thread (pairs with lag-based backpressure)
- `ParallelDispatcher`: one VT per record, unbounded fanout (pairs with in-flight backpressure)
- `KeyOrderedDispatcher`: per-key serial queue + per-key VT worker, in-flight backpressure across the consumer

`KPipeConsumer.processRecords` no longer branches on a boolean, it delegates to the dispatcher. In-flight ownership moves to each dispatcher (`SequentialDispatcher` doesn't need a counter at all; `ParallelDispatcher` and `KeyOrderedDispatcher` each own one). `totalInFlight()` now sums `dispatcher.pendingCount()` + buffered batch records.

### `KeyOrderedDispatcher` design

- LRU map keyed by record key (`LinkedHashMap` access-order, default cap 10,000 distinct keys, configurable via `withKeyOrderedMaxKeys(int)`).
- Each key gets an `ArrayDeque<Runnable>` and a `workerActive` flag, drained by a VT worker that exits when the queue empties. New records for the same key after the worker exits start a fresh worker.
- Eviction walks LRU oldest-first looking for an empty + idle queue. **Never evicts a non-empty queue**,  would break per-key ordering.
- If every queue at cap is non-empty (hot producer saturating distinct keys faster than they can drain), dispatch releases the lock + yields + retries. That stalls the consumer thread's poll loop, implicit backpressure for `KEY_ORDERED` mode.
- Null-keyed records all serialize through a single private sentinel object (`NULL_KEY`).

### API surface

```java
KPipe.json(topic, props)
  .pipe(/* operator */)
  .withProcessingMode(ProcessingMode.KEY_ORDERED)
  .withKeyOrderedMaxKeys(50_000)   // optional, default 10,000
  .toCustom(sink)
  .start();
```

Or via the explicit `KPipeConsumer.Builder`:

```java
KPipeConsumer.<String>builder()
  .withProperties(props)
  .withTopic(topic)
  .withPipeline(pipeline)
  .withProcessingMode(ProcessingMode.KEY_ORDERED)
  .build();
```

`withSequentialProcessing(boolean)` is deleted from both `KPipeConsumer.Builder` and the fluent `Stream` facade. All 23 callers (lib + tests + benchmarks + examples + docs) migrated to `withProcessingMode(ProcessingMode.SEQUENTIAL / .PARALLEL)` in this same commit.

### Offset semantics

Refactor is a dispatch-thread strategy change, **not** an offset-semantics change. `tryProcessRecord`, `markOffsetProcessed`, `handleProcessingError`, `KafkaOffsetManager`, the command queue, and the TrackOffset/MarkOffsetProcessed flow are byte-identical to pre-refactor. The lowest-pending-offset commit protocol works mode-agnostically — `ConcurrentSkipListSet.first()` handles arbitrary `MarkOffsetProcessed` arrival order by design.

## What's NOT in this PR

- Share-consumer benchmark, that's part 2 of #136, will land separately.

## Test plan

- [x] 12-test `KeyOrderedDispatcherTest` unit suite (same-key order, cross-key parallelism, null-key sentinel, LRU eviction of empty queues, worker liveness on throw, onComplete error swallow, pending counter semantics, etc.)
- [x] 12-test `DispatcherIntegrationTest` property suite, parameterized over all 3 `ProcessingMode` values via `MockConsumer`. Properties verified: liveness, strict per-partition offset order (`SEQUENTIAL`), strict per-key offset order (`KEY_ORDERED`), no concurrent processing (`SEQUENTIAL` + null-key serial queue), cross-key concurrency (`PARALLEL` + `KEY_ORDERED`), null-key serialization, and **offset-manager pending-set drain + commit-offset advancement across all 3 modes** (wires a real `KafkaOffsetManager`, asserts `pendingCount==0`, `highestProcessedOffset==N-1`, `nextOffsetToCommit==N` per partition).
- [x] 225 tests pass across all lib modules (`./gradlew :lib:*:check`); Docker-backed tests will run in CI.
- [x] `./gradlew spotlessCheck` clean.


eschizoid:
@chuck-alt-delete, friendly FYI I am going to be merging this bad boy likely this week  😄 

Let me know your thoughts when you have a chance.

PR #167, opened 2026-06-18
test(consumer): add focused unit tests for SequentialDispatcher and KeyOrderedDispatcher (closes #164)

### LINKED ISSUE #164: Add focused unit tests for SequentialDispatcher and KeyOrderedDispatcher
## Problem

`ParallelDispatcherTest` covers `ParallelDispatcher` lifecycle thoroughly: close-with-no-dispatch, dispatch-then-drain, interrupt-during-shutdown, reject-handler. The other two dispatchers in the §20 trio don't have an equivalent focused test.

## SequentialDispatcher

Pin:
- `pendingCount()` is 0 before dispatch, 1 during the inline run, 0 after.
- Exception thrown by `processTask` propagates (no swallow), `pendingCount()` still returns to 0.
- `close()` is idempotent.
- Behavior of dispatch after `close()` (inline-run anyway? reject? no-op?) — whichever the implementation guarantees, pin it.

## KeyOrderedDispatcher

Pin:
- LRU cap saturation logs WARNING **once** via the `AtomicBoolean.compareAndSet` guard — repeated saturation doesn't spam.
- Per-key serial ordering under concurrent dispatch on the same key (records for key K execute in dispatch order).
- `topKeyQueueDepths(n)` snapshot returns the deepest N queues by length.
- Worker-exits-when-queue-empties + new-record-spawns-fresh-worker handshake.
- Eviction skips non-empty queues; cap-saturated-all-non-empty case yields and retries (implicit backpressure).
- Null-keyed records share a single sentinel queue.
- `withKeyOrderedMaxKeys` builder warn-log when mode != KEY_ORDERED.

## Out of scope

`KeyOrderedDispatchBenchmark` already exists but measures throughput. Don't replace it — add a unit test alongside.

## Background

Flagged during PR #162 round-2 review.

eschizoid:
## Summary

Closes #164. The §20 dispatcher trio already had `ParallelDispatcherTest`, `SequentialDispatcherTest`, and `KeyOrderedDispatcherTest` in tree, but four of the invariants enumerated in the ticket weren't explicitly pinned. This PR closes those gaps without touching any production code.

### SequentialDispatcher (2 new tests)

- **`closeIsIdempotent`** — repeated `close()` calls must not throw.
- **`dispatchAfterCloseStillRunsInlineWithAccurateInFlight`** — pins the current "close is a no-op" behavior. SequentialDispatcher holds no resources to release; locking this in keeps a future change to reject post-close dispatch from being a silent semantic regression. Contrast: `ParallelDispatcher` rejects post-close dispatches, `KeyOrderedDispatcher` abandons on saturation-stall, `SequentialDispatcher` still runs.

### KeyOrderedDispatcher (2 new tests)

- **`saturationLogsWarningExactlyOnceAcrossRepeatedStalls`** — captures `System.Logger` output via the JUL bridge, drives two distinct saturation episodes against a `cap=2` dispatcher, and verifies the `AtomicBoolean.compareAndSet` guard fires the WARN exactly once. Regression guard against the log-flood footgun the one-shot was added to prevent.
- **`workerExitsWhenQueueEmptiesAndNewRecordSpawnsFreshWorker`** — proves the drain/dispatch handshake: the first worker VT exits after the queue drains (verified via `Thread.isAlive`), and a subsequent dispatch on the same key spawns a NEW worker (verified by thread-identity comparison), rather than silently enqueuing against a dead worker.

### Already covered by the pre-existing tests (no duplication)

The other ticket invariants are pinned in tree already and untouched here:
- `pendingCount` 0→1→0 across a Sequential dispatch.
- Exception propagation + in-flight rollback in Sequential.
- Per-key serial ordering invariant under concurrent dispatch.
- `topKeyQueueDepths(n)` ordering + cap + byte[] copy + null-key surfacing.
- Eviction skipping non-empty queues + cap-saturated stall + recovery.
- Null-keyed records sharing one sentinel queue.
- byte[] key normalization, defensive-copy, and post-dispatch mutation safety.
- `signalShutdown` releasing the saturation yield-loop, and not corrupting normal dispatch.
- `close()` interrupting stuck workers after the drain timeout.

### What is NOT pinned

None of the ticket's invariants were skipped. The "cap-saturated yields and retries" implicit-backpressure behavior is already covered by `dispatchStallUnderSaturationEventuallyRecoversWhenAQueueDrains` + `signalShutdownUnblocksStalledDispatchBeforeClose` in the existing test class, so this PR doesn't re-pin it.

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test --tests SequentialDispatcherTest --tests KeyOrderedDispatcherTest --tests ParallelDispatcherTest` — all green (5 + 24 + 3 = 32 tests).
- [x] Full `./gradlew :lib:kpipe-consumer:test` — 256 unit tests pass; one pre-existing Testcontainers flake (`KPipeProducerIntegrationTest > initializationError` — Kafka container `LogMessageWaitStrategy` timeout) is environmental, not related to this change.

### DIFF
--- a/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KeyOrderedDispatcherTest.java
+++ b/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KeyOrderedDispatcherTest.java
@@ -4,6 +4,7 @@ import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
@@ -19,6 +20,10 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #168, opened 2026-06-18
fix(consumer): document and assert single-writer invariant for commandQueue drain (closes #163)

### LINKED ISSUE #163: Fix non-atomic poll-then-readd race in drainCommandQueueForRevokedPartitions
## Problem

`KafkaOffsetManager.drainCommandQueueForRevokedPartitions` drains the command queue with a `for (var cmd = commandQueue.poll(); cmd != null; cmd = commandQueue.poll())` loop, filters revoked-partition entries out of each `CommitOffsets` command, and re-adds the survivors via `commandQueue.addAll(remainingCommands)`. **This sequence is not atomic.**

If anything writes a `CommitOffsets` to the command queue between the last `poll()` returning `null` and `addAll(remainingCommands)` completing, that new command jumps the queue and bypasses revoked-partition filtering. The consumer thread would then process it and commit against the new owner of the revoked partition — wrong-owner commit.

## Why it is quiet today

Only the consumer thread writes commit commands on the hot path, and `onPartitionsRevoked` runs on that same consumer thread. So no concurrent writer races against the drain in current code paths. But:

- The single-writer invariant is **not documented** and **not enforced**.
- A future contributor could add an off-thread `CommitOffsets` writer (ad-hoc commit from a different module, async commit hook, etc.) and silently break revoke correctness.

## Options

1. **Document + assert.** Add a Javadoc invariant on `commandQueue` and a debug-mode assertion that `onPartitionsRevoked` runs on the consumer thread. Cheapest. Doesn't fix the race, just guards against silently introducing it.
2. **Lock the drain.** Take a snapshot under a lock, swap the queue, process, restore. Real fix. Adds a lock on the rebalance path.
3. **Restructure.** Use a different concurrent primitive that supports atomic filter-in-place. Doesn't actually fix the intra-rebalance writer case — same race shape.

## Recommendation

Start with (1) — it makes the invariant explicit and trips a clear failure if anyone violates it. Upgrade to (2) if/when off-thread writers become a real need.

## Background

Flagged twice during PR #162 code review (silent-failure-hunter, code-reviewer) as pre-existing. Out of scope for that PR.

eschizoid:
## The race

`KafkaOffsetManager.drainCommandQueueForRevokedPartitions` drains `commandQueue` in a polling loop, filters revoked-partition entries out of each `CommitOffsets`, then re-adds the survivors via `commandQueue.addAll(remaining)`. The sequence is **not atomic** — anything writing a `CommitOffsets` between the last `poll()` returning `null` and `addAll(...)` completing would jump the queue and bypass revoked-partition filtering. The consumer thread would then commit against the new owner of the revoked partition.

## Why it is dormant today

The only `CommitOffsets` writer is `KafkaOffsetManager.performCommit`, called from the consumer thread, and `onPartitionsRevoked` runs on that same thread (dispatched by `Consumer.poll(...)`). So there is no concurrent writer racing the drain — but the single-writer invariant was neither documented nor enforced. A future contributor could quietly introduce an off-thread commit (async commit hook, ad-hoc commit from another module) and silently break revoke correctness.

## Why option (1) over option (2)

The ticket lists three options:

1. **Document + assert** — cheapest, doesn't fix the race, just guards against silently introducing it.
2. **Lock the drain** — real fix, adds a lock on the rebalance path.
3. **Restructure** — doesn't actually fix the intra-rebalance writer case.

Option (1) matches the "the race is theoretical today" reality. Option (2) would add a `ReentrantLock` taken on every commit enqueue *and* on the rebalance path for a race that no current writer can hit. The lock would also force every `performCommit` to contend with rebalance callbacks even though they're already serialized by being on the same thread. The right time to upgrade to (2) is when an off-thread writer becomes a real need — at which point this PR's assert will tell us loudly that the invariant has been broken.

## What changes

- **Javadoc on `commandQueue`** spells out the single-writer invariant: all writes happen from the Kafka consumer thread; `onPartitionsRevoked` runs on the same thread, so the drain is atomic with respect to writes. Explains exactly what kind of change would break it.
- **Javadoc on `drainCommandQueueForRevokedPartitions`** repeats the invariant at the method level so anyone reading the drain code sees why the poll-then-readd is safe today.
- **`assertOnConsumerThread()`** captures `Thread.currentThread()` on the first rebalance callback (stored in a `volatile Thread` field) and `assert`s subsequent callbacks run on that same thread. Bound to both `onPartitionsRevoked` and `onPartitionsAssigned`. `assert` is a no-op in production (no `-ea`) and trips loudly under Gradle's `Test` task (assertions on by default).
- **Two new tests** in `SingleWriterInvariantTests`:
  - `rebalanceCallbackFromDifferentThreadTripsAssertion` — binds the listener from the JUnit thread, then invokes from a different thread and expects an `AssertionError`.
  - `repeatedCallbacksOnSameThreadDoNotTrip` — calls all three callbacks from the same thread, expects no throw.

## Trade-off: first-touch capture

There is no clean seam for `KPipeConsumer` to hand the manager its consumer thread before `Consumer.poll(...)` first dispatches the listener. Adding a `registerConsumerThread(Thread)` method would expand the surface for one debug-mode check. Instead, the assertion treats "whoever ran first" as the source of truth and pins every subsequent caller to that thread. In production, that first call comes from the Kafka consumer thread.

## What would need to change if off-thread writers arrive

If we ever introduce an async commit hook (e.g., a metric-driven commit fired from a scheduled executor), the assert won't fire — it only checks the rebalance path. We would need to:

1. Upgrade to option (2): take a snapshot under a lock, swap the queue, process, restore.
2. Or refactor commit enqueue to go through a single point that the rebalance drain can coordinate with.

The Javadoc explicitly calls this out so the next contributor knows the failure mode they need to address.

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test --tests "io.github.eschizoid.kpipe.consumer.KafkaOffsetManagerTest"` passes (all 38 tests, including the 2 new ones)
- [x] Full `:lib:kpipe-consumer:test` passes except for one pre-existing Testcontainers / Docker init failure (`KPipeProducerIntegrationTest > initializationError`) unrelated to this change
- [ ] CI green

### DIFF
--- a/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/KafkaOffsetManager.java
+++ b/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/KafkaOffsetManager.java
@@ -60,13 +60,37 @@ public class KafkaOffsetManager<K> implements OffsetManager<K> {
 
   private final Consumer<K, byte[]> kafkaConsumer;
   private final AtomicReference<OffsetState> state = new AtomicReference<>(OffsetState.CREATED);
+
+  /// Shared command queue between `KPipeConsumer` and this manager.
+  ///
+  /// **Single-writer invariant.** All writes to this queue happen from the Kafka consumer thread
+  /// — `KPipeConsumer.consumerLoop` enqueues `Pause` / `Resume` / `Close` commands, and
+  /// `performCommit` enqueues `CommitOffsets`. The rebalance callback `onPartitionsRevoked`
+  /// (which calls `drainCommandQueueForRevokedPartitions`) is invoked by `Consumer.poll(...)`
+  /// on that same consumer thread, so the poll-then-readd drain in
+  /// `drainCommandQueueForRevokedPartitions` is atomic with respect to any writer.
+  ///
+  /// If a future change introduces an off-thread `CommitOffsets` writer (async commit hook, an
+  /// ad-hoc commit from another module, etc.), the drain becomes racy: a new command landing
+  /// between the last `poll()` returning `null` and `addAll(remaining)` completing would jump
+  /// the queue and bypass revoked-partition filtering — the consumer thread would then commit
+  /// against the new owner of a revoked partition. To enforce the invariant we capture the
+  /// consumer thread on the first listener callback and assert subsequent callbacks run on the
+  /// same thread (see `assertOnConsumerThread`).
   private final Queue<ConsumerCommand> commandQueue;
+
   private final Map<String, CompletableFuture<Boolean>> commitFutures = new ConcurrentHashMap<>();
   private final Map<TopicPartition, ConcurrentSkipListSet<Long>> pendingOffsets = new ConcurrentHashMap<>();
   private final Map<TopicPartition, Long> highestProcessedOffsets = new ConcurrentHashMap<>();
   private final Duration commitInterval;
   private volatile ScheduledExecutorService scheduler;
   private volatile ScheduledFuture<?> scheduledCommitTask;
+
+  /// Captured on the first rebalance callback to enforce the single-writer invariant documented
+  /// on `commandQueue`. `volatile` for visibility across the consumer thread and any test
+  /// thread that violates the invariant. Only ever written once: from `null` to the first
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

Issue #166, opened 2026-06-18
Audit examples/ — confirm fluent facade (KPipe.json/avro/...) is exercised, not just explicit Builder


eschizoid:
## Problem

§13 in CLAUDE.md marks the fluent facade (`KPipe.json/avro/protobuf/bytes/custom(...)` → `Stream<T>` → `Sink<T>` → `Handle`) as the **80% path** — the entry point we want most users on. But the `examples/` tree predates the facade and likely still demonstrates the explicit `MessageProcessorRegistry` + `KPipeConsumer.Builder` route.

If a new user follows the README into the examples, they may end up on the explicit path without realizing a simpler one exists.

## Audit

For each module under `examples/`:

- Does it use the fluent shape — `KPipe.X(topic, props).pipe(...).toCustom(...).start()` — or the explicit `KPipeConsumer.builder(...).withTopic(...).build()` shape?
- If explicit, is there a unique reason (escape-hatch demo: custom `OffsetManager`, custom Kafka consumer factory, custom rebalance hook, etc.) per §17?

## Goal

- **One** explicit example showcasing escape-hatch capabilities — make the reason it's explicit obvious from the example name and README intro.
- Everything else defaults to the fluent shape.
- README points to the fluent example as the canonical entry; the explicit one is labeled "advanced" or similar.

## Background

Flagged during PR #162 round-2 review. We market the fluent facade as the headline API; let's make sure the examples actually demonstrate it.

PR #183, opened 2026-06-18
docs(examples): add KPipe.bytes example (refs #166)

eschizoid:
## Summary

The audit in #166 flagged that `KPipe.bytes(topic, props)` was the only bundled format on the fluent facade without a corresponding `examples/` module. JSON, Avro, Protobuf, and the multi/demo paths all had complete reference apps; the raw-bytes path did not.

This PR adds `examples/bytes` to close that gap.

## What the example teaches

- `KPipe.bytes(topic, props).toCustom(MessageSink<byte[]>).start()` — the simplest possible end-to-end raw-bytes consumer.
- Demonstrates that the pipeline can stay in `byte[]` end-to-end with no format SerDe — useful for proxying, raw archival, or pre-format inspection where the consumer doesn't care about the payload structure.
- The custom `MessageSink<byte[]>` logs each record's length plus a short hex preview, mirroring what `bytesConsoleSink()` does internally but inlined so the reader can see how `toCustom(...)` plugs in.

## Structure

Copied from `examples/json/`:
- `examples/bytes/build.gradle.kts` — same shape as the json module, minus the JSON format dep (raw `byte[]` needs no format module — `MessageFormat.bytes()` lives in `kpipe-core`, transitively pulled by `kpipe-api`).
- `examples/bytes/src/main/java/io/github/eschizoid/kpipe/App.java` — `AppConfig.fromEnv()` + `KafkaConsumerConfig` + a 5-line `MessageSink<byte[]>` lambda.
- `settings.gradle.kts` — registers `examples:bytes`.

Refs #166 (does not close — that issue tracks broader audit follow-ups).

### DIFF
--- a/examples/bytes/build.gradle.kts
+++ b/examples/bytes/build.gradle.kts
@@ -1,0 +1,11 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
+tasks.named<ShadowJar>("shadowJar") {
+  manifest {
+    attributes(
+      "Main-Class" to "io.github.eschizoid.kpipe.App",
+    )
+  }
+}
+
+description = "KPipe - Kafka Consumer Application Using Raw Bytes"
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #184, opened 2026-06-18
docs(examples): add KPipe.custom-format example (refs #166)

eschizoid:
## Summary

Closes a gap surfaced by the #166 facade-audit: every other `KPipe.X(...)` entry point ships a runnable example except `KPipe.custom(topic, props, format)`. This PR adds `examples/custom-format`.

- **Format chosen: CSV-line, implemented inline.** `App.CsvOrderFormat` is ~20 LOC, demonstrates the entire `MessageFormat<T>` contract (`serialize`/`deserialize`), and intentionally throws on malformed input rather than returning `null` — null deserialization would mask decode failures as filtered records under the pipeline's `Result` semantics. CSV was picked over key=value pairs / fixed binary because the wire shape is obvious from the file format alone.
- **Why inline (not a new `kpipe-format-csv` module).** The point of `KPipe.custom(...)` is exactly that users can plug in any `MessageFormat<T>` from their own code without going through the registry or shipping a published module. Hoisting this CSV codec into `lib/kpipe-format-csv/` would defeat that demonstration — the example would just look like another `KPipe.json(...)` / `KPipe.avro(...)` clone.
- **Pipeline shape:** `KPipe.custom(topic, props, format).filter(qty > 0).pipe(uppercase sku).toCustom(loggingSink).start()`. Touches operator chaining + custom terminal sink so the fluent surface is visible end-to-end.
- **No extra deps.** Picks up `kpipe-api` + `kpipe-consumer` + `slf4j-simple` from `examples/build.gradle.kts`; no `kpipe-format-*` module reference.
- **Build:** `./gradlew :examples:custom-format:compileJava` and `:assemble` both pass locally. (Note: this worktree has a long-standing JGit/spotless-ratchet incompatibility unrelated to this change — `compileJava` was the verifying step.)

Registered in `settings.gradle.kts` next to the other example modules.

## Test plan

- [ ] `./gradlew :examples:custom-format:build` in a non-worktree checkout (spotless ratchet works there).
- [ ] Run the example against a local Kafka with a few CSV-shaped records (`id,sku,qty`) and confirm the sink logs the upper-cased SKU and the `qty <= 0` filter is applied.
- [ ] Confirm a malformed record (e.g. `foo,bar,not-a-number`) surfaces through the default error handler rather than getting silently dropped.

### DIFF
--- a/examples/custom-format/build.gradle.kts
+++ b/examples/custom-format/build.gradle.kts
@@ -1,0 +1,14 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
+tasks.named<ShadowJar>("shadowJar") {
+  manifest {
+    attributes(
+      "Main-Class" to "io.github.eschizoid.kpipe.App",
+    )
+  }
+}
+
+description = "KPipe - Kafka Consumer Application Using a User-Supplied MessageFormat"
+
+// No format-module dependency: the MessageFormat<T> is implemented inline in App.java to
+// demonstrate the user-supplied custom-format path through KPipe.custom(topic, props, format).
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #170, opened 2026-06-18
chore: inline Noop tracer/span singletons + delete dead RegistryFunctions methods (closes #165)

eschizoid:
## Summary

Executes two high-confidence cleanups identified by the architecture audit in #165.

### 1. Inline `NoopTracer` / `NoopSpanScope` as enum singletons

The two package-private classes
(`lib/kpipe-producer/.../tracing/NoopTracer.java`,
`NoopSpanScope.java`) existed only to hold `INSTANCE` constants reachable
through `Tracer.noop()` / `Tracer.SpanScope.noop()`. Collapsed them into
nested `enum Noop implements Tracer` / `enum Noop implements SpanScope` —
the idiomatic Java enum-singleton form — and deleted the two standalone
files. All 26 callsites reach the factory methods, so no callsite touches.

### 2. Delete dead `RegistryFunctions` methods + tests + fix doc rot

- Deleted `RegistryFunctions.createMetrics` — unreferenced in production,
  only `RegistryFunctionsTest` called it.
- Deleted `RegistryFunctions.withFunctionErrorHandling` — same, only the
  test referenced it.
- Removed the corresponding test cases.
- Fixed the class-level Javadoc that referenced a non-existent
  `withErrorHandling(fn, defaultValue)` overload — replaced with the
  actual `withOperatorErrorHandling(operator, logger)` signature.
- Verified `withConsumerErrorHandling` / `withOperatorErrorHandling` /
  `createUnmodifiableView` are still used by `MessageProcessorRegistry`
  and kept them.

## Net delta

`6 files changed, 39 insertions(+), 163 deletions(-)` — **−124 lines**.

## Behaviour change

Zero. Both factory seams (`Tracer.noop()`, `SpanScope.noop()`) preserved
verbatim. The deleted `RegistryFunctions` helpers had no production
callers; the audit's "dead in production" claim was verified by grepping
across the full repo before deletion.

## Test plan

- [x] `./gradlew :lib:kpipe-producer:test :lib:kpipe-core:test` — green
- [x] `./gradlew :lib:kpipe-consumer:test :lib:kpipe-tracing-otel:test --rerun-tasks` — green (verifies downstream tracing consumers compile + run against the inlined enums)
- [x] `./gradlew :lib:kpipe-api:test --rerun-tasks` — green

PR #185, opened 2026-06-18
test: pin three coverage follow-ups (onPartitionsAssigned, noop singleton-identity, WARN content)

eschizoid:
## Summary

Three small test-coverage follow-ups surfaced by recent code reviews. Each one closes a specific regression-detection gap with the lightest possible test.

- **Gap 1 (PR #168 review, criticality 5).** `KafkaOffsetManagerTest.SingleWriterInvariantTests` already covers the off-thread `onPartitionsRevoked` path but not the symmetric `onPartitionsAssigned` path. Added a sibling `partitionsAssignedFromDifferentThreadTripsAssertion` with identical shape: first same-thread call binds the consumer-thread reference, second call from a different thread must trip the `AssertionError`.
- **Gap 2 (PR #170 review, criticality 3).** After the `Tracer` / `SpanScope` enum inline, nothing pinned that `Tracer.noop() == Tracer.noop()` and `SpanScope.noop() == SpanScope.noop()`. A regression returning `new Noop()` instead of `INSTANCE` would silently allocate per-record on `KPipeConsumer`'s hot path. Added two `assertSame` lines to the existing `noopTracerInjectionAndExtractionAreSilentNoops` test in `OtelTracerTest` (the natural home given the test already exercises the noop branch).
- **Gap 3 (PR #167 review, criticality 4).** `saturationLogsWarningExactlyOnceAcrossRepeatedStalls` in `KeyOrderedDispatcherTest` asserted `count == 1` but not the message content. A reword that drops the "saturation" signal entirely would still pass. Added a loose lowercase `contains("saturat")` assertion — survives wording tweaks like "saturated" / "saturation" / "queues saturating" but catches a regression that genericizes the message.

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test :lib:kpipe-producer:test :lib:kpipe-tracing-otel:test` — all green
- [x] No new dependencies; only test-tree changes

### DIFF
--- a/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KafkaOffsetManagerTest.java
+++ b/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KafkaOffsetManagerTest.java
@@ -461,6 +461,37 @@ class KafkaOffsetManagerTest {
       );
     }
 
+    /// Mirror of [#rebalanceCallbackFromDifferentThreadTripsAssertion] for the
+    /// `onPartitionsAssigned` half of the listener. The first same-thread call binds the consumer
+    /// thread; the second call from a different thread must trip the invariant.
+    @Test
+    void partitionsAssignedFromDifferentThreadTripsAssertion() throws Exception {
+      final var listener = offsetManager.createRebalanceListener();
+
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #186, opened 2026-06-18
test(consumer): pin ParallelDispatcher pendingCount + onComplete race paths

eschizoid:
## Summary

Closes the criticality-9 coverage gap on `ParallelDispatcher`. The dispatcher's `AtomicLong inFlight` is the single source of truth for PARALLEL-mode backpressure (see CLAUDE.md §5 and §20). A leak (missing decrement) silently latches the consumer at HIGH watermark → pause-forever deadlock. A double-decrement underflows toward negative → pause-never overload. Neither is visible from logs — both are silent-failure modes worth pinning with explicit tests.

Pre-existing coverage in `ParallelDispatcherTest` covered the happy path and the close-interrupts-running-task path. Four new tests close the gaps around `processTask` throwing and concurrent close.

## Invariants pinned

- **`pendingCountDecrementedWhenTaskThrows`** — a throwing task body still drains the in-flight counter. Catches the failure mode where a `try { processTask.run(); } finally { inFlight.decrementAndGet(); }` was ever refactored into a `try/catch` that skipped the finally on the exception path.
- **`onCompleteFiresExactlyOnceWhenTaskThrows`** — `onComplete` fires exactly once on the throw path. The consumer's `afterRecordComplete()` unparks the consumer thread when backpressure is held (§11); zero invocations stall the pipeline on the first exception, double invocations spuriously double-wake.
- **`dispatchAfterCloseHitsRejectHandler`** — post-`close()` dispatches route to the reject handler, the task body does NOT execute, `onComplete` is NOT invoked (the consumer takes its reject path instead), and the speculative `inFlight.incrementAndGet()` in `dispatch()` is rolled back. Without rollback, every shutdown-race rejection would leak a count and `shutdownGracefully()` would burn its full timeout waiting on a ghost.
- **`concurrentDispatchAndCloseDoesNotLeakPending`** — stress: 200 concurrent dispatches racing against `close()`. Asserts `pendingCount() == 0` after settle AND `started + rejected == total` (no silent drops). A third "leaked" state shows up either as non-zero `pendingCount` or as `started + rejected < total`.

## Style notes

- Virtual-thread tests use `Thread.ofVirtual()` + `CountDownLatch` per project conventions (no `CompletableFuture`).
- `///` Javadoc, `final var` locals, `System.Logger` only — consistent with the existing 3 tests in the file.
- No production code touched — pure test addition.

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test --tests "io.github.eschizoid.kpipe.consumer.ParallelDispatcherTest"` — 7/7 green
- [x] CI green on PR
- [x] Copilot review pass with no new comments

### DIFF
--- a/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/ParallelDispatcherTest.java
+++ b/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/ParallelDispatcherTest.java
@@ -65,6 +65,152 @@ class ParallelDispatcherTest {
   }
 
   @Test
+  void pendingCountDecrementedWhenTaskThrows() throws InterruptedException {
+    // Guards the criticality-9 race in §20: ParallelDispatcher owns the in-flight counter that
+    // drives PARALLEL-mode backpressure (§5). If the task body throws and the finally block did
+    // NOT decrement, pendingCount() would climb monotonically per failed record — the watermark
+    // would latch at HIGH and the consumer would pause forever (deadlock). Conversely, double
+    // decrement would underflow toward negative and the consumer would never pause (false-
+    // negative backpressure). The production code wraps processTask.run() in try/finally so
+    // interrupt-style and exception-style exits both decrement exactly once; this test pins it.
+    final var rejectCount = new AtomicInteger(0);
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #188, opened 2026-06-18
test(consumer): pin BatchPipelineWrapper coverage-contract paths

eschizoid:
## Summary

Adds 5 tests to `BatchPipelineWrapperCoverageContractTest` that pin the silent-failure guards in `BatchPipelineWrapper`. The existing `BatchPipelineWrapperConcurrencyTest` only exercises the happy-path parallel-enqueue invariant — these tests close the remaining gap on the wrapper's `BatchResult` classification and drain logic.

The coverage contract is the wrapper's last line of defense against a buggy `BatchSink` silently marking records as processed that the sink never confirmed. Together, `BatchResult.succeededIndexes` and `BatchResult.failedByIndex.keys()` must cover every input position; the wrapper flags missing indexes with a synthetic `IllegalStateException` and routes them to the failure callback (DLQ) rather than dropping them on the floor.

## Tests added

| # | Test | Invariant pinned | Silent-failure mode caught |
|---|------|------------------|----------------------------|
| 1 | `incompleteBatchResultRoutesMissingRecordsToDlq` | A `BatchResult` whose union doesn't cover `[0, size)` must route the missing positions to `onBatchFailure` with a synthetic `IllegalStateException` cause. | A buggy sink returning `BatchResult` covering N-1 of N indexes would otherwise leave the N-th record looking "processed" without the sink ever confirming the write — exactly the "processed counter rises while sink invocations stay at 0" mode the wrapper was hardened against. |
| 2 | `outOfRangeIndexInBatchResultLogsWarning` | An out-of-range succeeded/failed index logs WARNING and falls through to the coverage check so real records still hit the failure callback. | Without the WARN, an operator has no breadcrumb when a sink emits garbage indexes (e.g. `99` on a batch of 3); without the fall-through, an out-of-range index could mask a missing real index. |
| 3 | `nullBatchResultTreatedAsWholeBatchFailure` | A null `BatchResult` is treated as whole-batch failure — every record routed to `onBatchFailure` with the same shared synthetic cause. | A `null` return would otherwise NPE inside the coverage check and silently kill the flush thread, freezing the batch pipeline. |
| 4 | `ageTickFlushesBeforeSizeThreshold` | The scheduler age tick flushes a single buffered record when the size threshold isn't met. | Without the age trigger, a slow producer's last record could sit in the buffer forever — never flushed, never committed. |
| 5 | `closeFlushesBufferedRecordsBeforeOffsetManagerClose` | `close()` drains the buffer and invokes `markProcessed` for every buffered record before returning. | This is the wrapper-level guarantee that lets `KPipeConsumer.releaseConstructedResources` call `offsetManager.close()` afterwards with all batch offsets accounted for. Without it, a graceful shutdown would lose the last in-buffer records' offsets and they'd be re-fetched on restart. |

## Notes

- Tests live in `io.github.eschizoid.kpipe.consumer` package (matching the existing `BatchPipelineWrapperConcurrencyTest`) so the package-private `BatchPipelineWrapper` constructor is accessible. The bench harness (`BatchSinkLatencyBenchmark`) drives through `MockConsumer` because it lives outside the package — same-package tests can construct the wrapper directly, which is cleaner here.
- Test #4 uses two latches: `sinkLatch` fires when the sink is invoked, and a separate `markProcessedLatch` fires when the wrapper has finished calling `markProcessed`. The wrapper invokes the callback *after* the sink returns, so a sink-only latch races with the callback-side assertion (caught during full-suite runs of an earlier draft).
- Test #5 records relative ordering between callback events and a simulated `OffsetManager.close()` to pin the ordering invariant at the wrapper level; the actual production sequence is enforced in `KPipeConsumer.releaseConstructedResources` (`dispatcher.close()` → batch wrappers `close()` → `offsetManager.close()`).
- No bugs were found in the production code — every test passed first try (the one initial failure was a test-side race, not a wrapper bug).

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test --tests "io.github.eschizoid.kpipe.consumer.BatchPipelineWrapperCoverageContractTest"` — 5/5 pass
- [x] `./gradlew :lib:kpipe-consumer:test` — full suite green (265 tests)
- [ ] CI passes

### DIFF
--- a/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapperCoverageContractTest.java
+++ b/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapperCoverageContractTest.java
@@ -1,0 +1,366 @@
+package io.github.eschizoid.kpipe.consumer;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.github.eschizoid.kpipe.sink.BatchPolicy;
+import io.github.eschizoid.kpipe.sink.BatchResult;
+import io.github.eschizoid.kpipe.sink.BatchSink;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #191, opened 2026-06-18
test(producer): pin Tracer header injection through DLQ

eschizoid:
## Summary

Closes a criticality-5 coverage gap: a future refactor that drops the `tracer.injectContextInto(headers)` call in `KPipeProducer.sendToDlq` would silently strip trace context from DLQ records without any existing test failing.

- Adds `dlqRecordCarriesTracerInjectedHeaders` to `KPipeProducerTest`.
- Wires a tiny anonymous `Tracer` stub (not Mockito) that stamps `x-tracer-stamp: yes` during `injectContextInto`.
- Asserts the stamp lands on the captured outbound `ProducerRecord`'s headers after a `sendToDlq` call.

Existing `shouldSendDlqRecordWithAllHeaders` covers the static `x-dlq-*` enrichment headers but never exercises the tracer injection branch; using a hand-written stub keeps the contract enforced by behaviour rather than by Mockito argument matching.

No production-code change. Production already invokes the tracer correctly at `KPipeProducer.java:183` — this PR just pins that invariant under test.

## Test plan

- [x] `./gradlew :lib:kpipe-producer:test` — green (24 tests in `KPipeProducerTest`, was 23).
- [x] New test exercises a real anonymous `Tracer` implementation, not a mock.
- [x] No production code touched.

### DIFF
--- a/lib/kpipe-producer/src/test/java/io/github/eschizoid/kpipe/producer/KPipeProducerTest.java
+++ b/lib/kpipe-producer/src/test/java/io/github/eschizoid/kpipe/producer/KPipeProducerTest.java
@@ -14,6 +14,7 @@ import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #193, opened 2026-06-18
test(schema-registry): pin CachedSchemaResolver failure-caching contract (§19)

eschizoid:
## Summary

Closes the criticality-8 coverage gap in `CachedSchemaResolver` flagged
by the prior coverage audit: the failure-propagation path under
contention. The §19 "cache forever in-process" invariant in
`CLAUDE.md` only holds for **successful** lookups — a failed delegate
call must NOT poison the cache with a sentinel, or a transient SR
outage would pin a topic into a permanent decode-error state.

No production change. The current implementation relies on the
`ConcurrentHashMap.computeIfAbsent` contract (mapping-function throws →
no entry stored, exception propagates) and these tests now lock that in.

## Tests

- **`failedLookupIsNotCachedAndIsRetriedOnNextCall`** — delegate throws
  on the first call, succeeds on the second. Asserts the exception
  propagates verbatim, `cache.size()` stays 0 after the failure, and the
  retry actually re-invokes the delegate (proving no sentinel). Catches:
  a regression where `computeIfAbsent` was wrapped in a try/catch that
  silently stored an empty-string sentinel — the second call would
  either also throw or return `""` without hitting the delegate.

- **`concurrentMissOnSameIdCollapsesToOneDelegateCall`** — two virtual
  threads race a miss on the same id; the first is parked inside the
  delegate via a `CountDownLatch` so the second is guaranteed to enter
  `computeIfAbsent` while the bucket is held. Asserts exactly one
  delegate call and both threads see the same schema string. Catches: a
  regression to a check-then-act `get` / `put` shape that would let both
  threads issue the HTTP call.

- **`concurrentFailureOnSameIdDoesNotPoisonCache`** — 16 virtual threads
  race a failing delegate, then one recovery call with a successful
  delegate. Asserts every racing thread observes the failure, `size()`
  is 0 through the storm, the recovery call succeeds and now caches.
  Catches: a regression that races a cache-write before the delegate
  throw clears state (e.g. a misplaced `cache.put(id, ...)` outside the
  `computeIfAbsent` mapping function).

## §19 wire-up

Per the §19 invariants in `.claude/CLAUDE.md`:

> `computeIfAbsent` atomicizes load+store so concurrent misses on the same ID collapse to one HTTP call.

Test #2 is the direct executable proof of that line. Tests #1 and #3
are the implicit corollary: the "cache forever" claim is conditional on
"...after a successful resolve" — anything else (failure storms, partial
state) must leave the cache empty.

## Test plan

- [x] `./gradlew :lib:kpipe-schema-registry-confluent:test` — 11 tests pass (8 prior + 3 new), 0 failures.
- [x] All three new tests use virtual threads via `Thread.ofVirtual()` + `CountDownLatch` per project test doctrine.
- [x] No production change; the existing `CachedSchemaResolver` already satisfies the contract.

### DIFF
--- a/lib/kpipe-schema-registry-confluent/src/test/java/io/github/eschizoid/kpipe/schemaregistry/confluent/CachedSchemaResolverTest.java
+++ b/lib/kpipe-schema-registry-confluent/src/test/java/io/github/eschizoid/kpipe/schemaregistry/confluent/CachedSchemaResolverTest.java
@@ -130,8 +130,8 @@ class CachedSchemaResolverTest {
     /// If the cache poisoned itself with a sentinel on failure, the second call would either
     /// throw again (sentinel re-raised) or skip the delegate entirely (sentinel served as a
     /// silent success). Either way the topic would be pinned into a permanent decode-error
-    /// state, violating the §19 "cache forever in-process" invariant which only applies AFTER a
-    /// successful lookup.
+    /// state. Confluent SR schema IDs are immutable, so the "cache forever" rule only applies
+    /// AFTER a successful lookup — failures must propagate without leaving residue.
     final var calls = new AtomicInteger();
     final var shouldFail = new AtomicReference<>(Boolean.TRUE);
     final SchemaResolver flaky = id -> {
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #197, opened 2026-06-19
chore(test-fixtures): inline triplicated CapturingHandler into per-module *ConsoleSinkTest

eschizoid:
## Summary

The same 30-line `java.util.logging.Handler` lived verbatim in three places, each with exactly one consumer (its sibling `*ConsoleSinkTest`):

- `lib/kpipe-format-avro/.../CapturingHandler.java` → consumed by `AvroConsoleSinkTest`
- `lib/kpipe-format-json/.../CapturingHandler.java` → consumed by `JsonConsoleSinkTest`
- `lib/kpipe-format-protobuf/.../CapturingHandler.java` → consumed by `ProtobufConsoleSinkTest`

The three files were byte-for-byte identical except for the package declaration: raw `record.getMessage()` capture into a `ByteArrayOutputStream`, no formatting, no parameter substitution, no separator.

## What changed

- Inlined `CapturingHandler` as a `private static final class` nested at the bottom of each sole-consumer test, mirroring how `lib/kpipe-api/src/test/java/io/github/eschizoid/kpipe/KPipeBytesConsoleSinkTest.java` already structures its (slightly richer) handler.
- Deleted the three top-level `CapturingHandler.java` files.
- Added a brief `///` Javadoc on each inlined class so future readers know it is the "raw, no-formatting" flavour.

## Why not a shared `kpipe-test-fixtures` module?

A shared module would be over-engineering for a 30-line JUL handler:

- It would introduce a new published Maven coordinate (or test-only artifact + per-module dependency wiring) just to host one class.
- The split-package rule means the fixture would need its own package (e.g. `io.github.eschizoid.kpipe.testfixtures`), so the three call sites would need an `import` change anyway — the reuse cost is no smaller than copy-and-trim.
- The handler has zero domain dependencies and is genuinely simple; the duplication had no maintenance burden so far (no drift, no bug fixes ever needed).
- Each test's nested-class flavour can now diverge if needed (e.g. parameter substitution, separators) without touching siblings.

Inline is the right call here.

## Net LOC

`6 files changed, 87 insertions(+), 87 deletions(-)` — break-even on raw line count, but three identical top-level definitions collapse to zero, and each handler is now scoped to its sole user.

## Test plan

- [x] `./gradlew :lib:kpipe-format-avro:test :lib:kpipe-format-json:test :lib:kpipe-format-protobuf:test` → BUILD SUCCESSFUL
- [x] `grep -rn "CapturingHandler" lib/ examples/ benchmarks/ --include="*.java"` shows no more top-level `CapturingHandler.java` files, only the three private nested classes plus the pre-existing `KPipeBytesConsoleSinkTest` nested class (which was already private and is not touched here).


eschizoid:
@copilot please review

PR #198, opened 2026-06-19
chore(tracing-otel): inline HeadersTextMap{Getter,Setter} into OtelTracer

eschizoid:
## Summary

- Both `HeadersTextMapGetter` and `HeadersTextMapSetter` were package-private SPI adapters with exactly one production callsite (`OtelTracer` itself), zero tests, zero examples. Their Javadoc already declared them as "Used by `OtelTracer#...`" — conceptually private helpers leaking out as top-level types.
- Inlined both as `private static final class` nested inside `OtelTracer` (kept as parallel nested classes rather than mixing a lambda for the setter and a class for the getter — `TextMapGetter` has two methods so it can't be a lambda anyway, and the symmetric shape reads better).
- Preserved the stateless-singleton `INSTANCE` semantics from the originals; callsites were updated from `HeadersTextMapGetter.INSTANCE` / `HeadersTextMapSetter.INSTANCE` to the nested `HeadersGetter.INSTANCE` / `HeadersSetter.INSTANCE`.
- Dropped the `[HeadersTextMapSetter] / [HeadersTextMapGetter]` bullet from `package-info.java` and folded the propagation note into the `OtelTracer` bullet.

## Cohesion win

- The adapters are now invisible outside `OtelTracer` — they cannot be referenced from anywhere else in the module, which matches their actual usage. One fewer top-level surface for `kpipe-tracing-otel` to defend.
- Two fewer compilation units in the module (3 → 1 production class).
- LOC delta: **-61 / +53 = -8 lines** including the Javadoc carry-over and the new imports for `TextMapGetter` / `TextMapSetter` / `Header` / `ArrayList` / `StandardCharsets` in `OtelTracer`.

## JPMS

- `module-info.java` only `exports io.github.eschizoid.kpipe.tracing.otel;` (whole package). The deleted classes were package-private, so they were never independently visible. No module-info edit needed.

## Test plan

- [x] `./gradlew :lib:kpipe-tracing-otel:test` → BUILD SUCCESSFUL
- [x] `grep -rn 'HeadersTextMapGetter\|HeadersTextMapSetter' lib/ examples/ benchmarks/ --include='*.java' --include='*.kt'` returns no hits


eschizoid:
@copilot please review

PR #199, opened 2026-06-19
fix(api): MultiBuilder rejects silently-dropped per-route consumer-wide settings

eschizoid:
## Summary

`MultiBuilder.start()` was a §12 silent-failure: per-route consumer-wide
settings like `withRetry`, `withDeadLetterTopic`, `withErrorHandler`,
`withPollTimeout`, `withMetrics`, `withTracer`, `withCircuitBreaker`,
and `withBackpressure` were quietly dropped when a route configurator
set them. The previous `rejectPerRouteConsumerWideSettings` guard only
caught `withProcessingMode` + `withKeyOrderedMaxKeys`; everything else
fell on the floor.

### The bug

`MultiBuilder.start()` only invokes `ds.buildPipeline()` /
`addBatchRoute(...)` for each route. `DefaultStream.applyCommonConsumerConfig`
(which propagates retry / backpressure / metrics / errorHandler /
dlqTopic / pollTimeout / tracer onto the underlying `KPipeConsumer.Builder`)
is **never called** from the multi path. Every per-route consumer-wide
setting other than the two already guarded was silently lost.

### Reproduction (before this PR)

```java
KPipe.multi(props)
  .json("t", s -> s.withRetry(3, Duration.ofSeconds(1)).withDeadLetterTopic("dlq"))
  .start();
```

…built a consumer with **neither retry NOR DLQ** wired, and emitted no
error. The misconfig surfaced only as missing behavior at runtime —
exactly the overloaded-null / silent-drop footgun §12 exists to prevent.

### Fix

Extend `MultiBuilder.rejectPerRouteConsumerWideSettings` to throw
`IllegalArgumentException` when a per-route configurator sets any
consumer-wide value. Two message shapes:

- **Has a symmetric `MultiBuilder.with*` mirror** (`withMetrics`,
  `withTracer`, `withCircuitBreaker`): error points users at the
  consumer-level setter.
- **No mirror yet** (`withRetry`, `withBackpressure`,
  `withDeadLetterTopic`, `withErrorHandler`, `withPollTimeout`):
  error points users at the explicit `KPipeConsumer.Builder`
  escape hatch and notes the deferred multi-builder mirror follow-up.

### Deferred follow-up

Adding consumer-wide mirrors on `MultiBuilder` for `withRetry` /
`withDeadLetterTopic` / `withErrorHandler` / `withPollTimeout` /
`withBackpressure` is a separate change — they need design decisions
(e.g. should DLQ be a consumer-wide single topic or per-route routing?).
This PR's rejection message documents the gap so users hit a loud error
with a workable next step, instead of a silently-broken consumer.

## Test plan

- [x] One test per rejected setting: `withRetry`, `withBackpressure`,
  `withDeadLetterTopic`, `withErrorHandler`, `withPollTimeout`,
  `withMetrics`, `withTracer`, `withCircuitBreaker`. Each asserts the
  thrown `IllegalArgumentException` names the offending setter and the
  offending route topic, and (for mirrored settings) the message points
  at the consumer-level `MultiBuilder.with*` setter.
- [x] Smoke test that the consumer-wide `MultiBuilder` setters
  (`withMetrics`, `withTracer`, `withCircuitBreaker`) still build
  cleanly together with the existing `withProcessingMode` /
  `withKeyOrderedMaxKeys` chain.
- [x] `./gradlew :lib:kpipe-api:test` — all 21 cases in
  `KPipeFacadeBuildTest` pass; full module suite green.

### DIFF
--- a/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/MultiBuilder.java
+++ b/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/MultiBuilder.java
@@ -242,13 +242,14 @@ public final class MultiBuilder {
     return DefaultHandle.startAndWrap(consumerBuilder.build());
   }
 
-  /// Processing mode and the KEY_ORDERED LRU cap are consumer-wide settings — one
-  /// [KPipeConsumer] runs in exactly one mode with one cap. If a route configurator sets a
-  /// non-default value (`s -> s.withProcessingMode(...)` or
-  /// `s -> s.withKeyOrderedMaxKeys(...)`), the per-route setting would be silently dropped
-  /// because [#start()] builds a single consumer. Detect that here and fail loudly so the
-  /// user moves the call up to `MultiBuilder.withProcessingMode(...)` /
-  /// `MultiBuilder.withKeyOrderedMaxKeys(...)`.
+  /// Consumer-wide settings live on one [KPipeConsumer] (one consumer-group, one offset
+  /// manager, one poll loop), so a per-route configurator can only ever set them for itself —
+  /// [#start()] then drops them on the floor when it folds N routes into a single consumer.
+  /// Detect every such case here and fail loud so the misconfig surfaces at construction time
+  /// instead of as a silent missing-retry / missing-DLQ at runtime.
+  ///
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #203, opened 2026-06-19
perf: hot-path quick wins — hoist getSimpleName, batch scratch buffer + BitSet, drop KeyOrdered wrapper

eschizoid:
## Summary

Three small, surgical allocation cuts on the per-record hot path. No behavior change; existing tests pass unmodified.

### 1. `TypedPipelineBuilder.deserialize` — hoist `getSimpleName()`

`format.getClass().getSimpleName()` was recomputed on every `deserialize()` call, but the value is only used inside the catch arm (diagnostic message wrap). Moved it to a `final` capture in `build()` so it's computed exactly once per built pipeline.

**Allocation eliminated:** one `String` plus a class-hierarchy walk per record on the success path.

### 2. `BatchPipelineWrapper.flushLocked` — scratch buffer reuse + `BitSet`

- Replaced `List.copyOf(buffer)` + separate `ArrayList<T>` for values with one pre-sized `ArrayList<>(buffer)` snapshot and one pre-sized values list. The `buffer` field is now reused (`ArrayList.clear()` keeps the backing array), so steady-state flushes don't reallocate the buffer.
- Replaced `HashSet<Integer>(size * 2)` for `succeeded` with `BitSet(size)` — constant-time `set` / `get` on primitive `int` indexes, zero `Integer` boxing on the walks.
- Replaced `ArrayList<Integer> missing` with a count-first pass; only allocate the list on the (rare) coverage-violation path. Common case is allocation-free for the missing-index tracking.

**Allocations eliminated per flush:** the immutable `List.copyOf` array, the `HashSet` plus its boxed `Integer` entries, the always-allocated `missing` list, and the buffer's backing array on every steady-state flush.

### 3. `KeyOrderedDispatcher.dispatch` — drop the wrapper `Runnable`

The dispatcher used to wrap `(processTask, onComplete)` in a per-dispatch lambda that also captured `pending` and `key`. Replaced with a small `QueuedTask` record holding only `(processTask, onComplete)`. `drain()` now handles `pending.decrement` + `onComplete` directly in its own try/finally, preserving the same throw-safety + drain-loop-survives-any-`Throwable` semantics as the old wrapper.

The dispatcher's public `dispatch(record, processTask, onComplete)` signature is unchanged — all existing dispatcher tests pass without modification.

**Allocation eliminated:** the wrapper `Runnable` (captures 4 things) becomes a 2-field record. Lower object header + capture overhead per dispatch.

## Deferred for JMH-baselined PRs

Bigger wins surfaced by the same audit are deliberately not in this PR — they need before/after benchmark numbers, not just "looks faster":

- `OffsetManager`: cache `TopicPartition` instances per `(topic, partition)` instead of allocating one per record's `markOffsetProcessed` / `trackOffset`.
- `MessageFormat` SPI: add `deserialize(byte[], int, int)` so the `skipBytes(n)` path can skip the `System.arraycopy` on every record.

## Test plan

- [x] `./gradlew :lib:kpipe-core:test` — green (change 1 covered)
- [x] `./gradlew :lib:kpipe-consumer:test` — green twice, 263 unit tests pass on both runs. Two `@Testcontainers` integration tests (`ExternalOffsetIntegrationTest`, `KPipeProducerIntegrationTest`) fail with `DockerClientProviderStrategy` errors — pre-existing failure on `main` in any environment without a running Docker daemon, unrelated to these changes.
- [x] All `KeyOrderedDispatcherTest` cases pass without modification (signature preserved).
- [x] All `BatchPipelineWrapper*Test` cases pass — coverage contract, concurrency, and out-of-range index handling all green.

### DIFF
--- a/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapper.java
+++ b/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapper.java
@@ -7,7 +7,7 @@ import io.github.eschizoid.kpipe.sink.BatchSink;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.BitSet;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #205, opened 2026-06-20
perf(consumer): cache TopicPartition per partition in KafkaOffsetManager + add JMH benchmark

eschizoid:
## Summary

`KafkaOffsetManager.{trackOffset, markOffsetProcessed}` previously allocated a fresh `TopicPartition(topic, partition)` per call to use as a `ConcurrentHashMap` key. At 100k rec/s that's ~200k disposable `TopicPartition` instances/sec from this manager alone — twice per record (consumer thread + worker thread).

This PR introduces a per-partition cache so the per-record allocation is gone, plus a JMH benchmark to measure it.

## The cache

- **Shape:** `ConcurrentHashMap<String, ConcurrentHashMap<Integer, TopicPartition>>` (outer keyed by topic name, inner by partition number).
- **Lookup:** `outerMap.computeIfAbsent(topic, _ -> new ConcurrentHashMap<>()).computeIfAbsent(partition, p -> new TopicPartition(topic, p))`. Fast path = two CHM lookups returning a stable instance.
- **Eviction:** the existing rebalance listener already iterates `onPartitionsRevoked` partitions to clear `pendingOffsets` and `highestProcessedOffsets`. The cache eviction is added in the same loop. `onPartitionsLost` delegates to `onPartitionsRevoked` so it gets the same treatment.
- **Bucket-locked removal:** `evictCachedTopicPartition` uses `computeIfPresent` on the outer map so removing a partition and dropping the (now-empty) inner map happens under one bucket lock — a concurrent `cachedTopicPartition` call for a sibling partition under the same topic cannot see a transient empty inner map and double-allocate.

## Out of scope (deferred)

- **`Long` boxing** in `ConcurrentSkipListSet<Long>.add/remove` and the `compute*` lambdas — that's a data-structure change (would need a primitive-long sorted set). Separate PR.
- **The `Math.max` lambda** in `compute(... Math.max ...)` — JIT typically inlines the single call site; not worth a behavioral change for a sub-measurable win.

## Tests

- All existing `KafkaOffsetManagerTest` cases pass unchanged — behavior is identical.
- New `TopicPartitionCacheTests` nested class verifies:
  - The cached `TopicPartition` is `assertSame` across multiple track/mark calls for the same partition.
  - Revoked partitions are evicted from the cache after `onPartitionsRevoked`.
  - A fresh `TopicPartition` (`assertNotSame` to the original) is allocated after re-assignment of a previously-revoked partition.
  - Evicting one partition under a topic preserves cached siblings.

## Benchmark

`OffsetManagerBoxingBenchmark` drives `trackOffset` + `markOffsetProcessed` through the public API with pre-built synthetic `ConsumerRecord`s. Cache is pre-warmed in `@Setup` so the timed loop only exercises the fast path.

### Smoke run (single-shot, post-fix only — not a baseline)

```
# JMH 1.36, JDK 25.0.3 OpenJDK 64-Bit Server VM
# Warmup: 1 iteration, 2s; Measurement: 1 iteration, 3s; Fork: 1; Threads: 1

Benchmark                                  (partitions)   Mode  Cnt        Score   Error  Units
OffsetManagerBoxingBenchmark.trackAndMark             8  thrpt       3955634.128          ops/s
```

This is the cached path only — single-shot run intended just to verify the bench compiles and produces sensible numbers. **No baseline comparison is asserted here.** Run with full iteration counts plus `-prof gc` to capture the real before/after; the `gc.alloc.rate.norm` column is the primary signal (one `TopicPartition` instance ~= 24 bytes on a 64-bit JVM with compressed oops should disappear per record).

Reproduce:

```bash
./gradlew :benchmarks:jmh \
  -Pjmh.includes='OffsetManagerBoxingBenchmark' \
  -Pjmh.profilers='gc'
```

## Test plan

- [x] `./gradlew :lib:kpipe-consumer:test --tests "*KafkaOffsetManagerTest*"` — green
- [x] `./gradlew :benchmarks:compileJmhJava` — green
- [x] Smoke-run `OffsetManagerBoxingBenchmark` with `-i 1 -wi 1 -f 1` — produces sensible numbers
- [ ] Reviewer runs full JMH with `-prof gc` against `main` baseline + this branch to confirm allocation-rate drop

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/kpipe/benchmarks/OffsetManagerBoxingBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/kpipe/benchmarks/OffsetManagerBoxingBenchmark.java
@@ -1,0 +1,143 @@
+package io.github.eschizoid.kpipe.benchmarks;
+
+import io.github.eschizoid.kpipe.consumer.ConsumerCommand;
+import io.github.eschizoid.kpipe.consumer.KafkaOffsetManager;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.TimeUnit;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.MockConsumer;
+import org.apache.kafka.common.TopicPartition;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #207, opened 2026-06-20
fix(format): enrich SerDe error messages with format name + byte context

eschizoid:
## Why

The no-silent-failures doctrine exists because a wrong `skipBytes(5)` once made protobuf deserialization fail while a "processed" counter kept rising. The doctrine cares about the **signal** when deserialize genuinely throws — and that signal is the message + cause chain. The bare `"Failed to deserialize X"` messages buried the format name and byte context behind the nested cause, so an operator triaging a log line couldn't tell **which** format, **how many** bytes, or **what mode** without unwrapping the stack trace.

## What

- **JsonFormat.deserialize** — message now reports input byte length + a short leading-byte hex preview. `serialize` reports the map entry count.
- **ProtobufFormat.deserialize** — message reports byte length, `descriptor.getFullName()`, leading-byte hex preview, and a Confluent-envelope `skipBytes(6)` hint.
- **AvroFormat** — static-mode deserialize message reports byte length + `staticSchema.getFullName()`, matching the quality of the existing envelope path. `serialize` reports the record's schema name. Static deserialize now also catches Avro's own `RuntimeException` decode failures (e.g. "Malformed data. Length is negative") so the enriched message wins instead of the bare buried one — previously the most common malformed-bytes failure escaped uncaught.

## Constraints honored

- Exception **type** unchanged — still `RuntimeException` wrapping the original cause. Only the message changed; callers catching `RuntimeException` are unaffected.
- Hex-preview helper caps at 8 bytes and handles empty/short arrays without throwing (a deserialize error on a 0- or 2-byte payload still produces a clean message, no AIOOBE).
- Messages are single-line (no embedded newlines) for log friendliness.

## Tests

Each format gains a malformed-bytes test asserting the thrown message contains the format name AND the byte length, pinning the triage-signal contract. `JsonFormatTest` is new (JSON previously had only pipeline/registry tests); Avro and Protobuf tests are extended.

`./gradlew :lib:kpipe-format-json:test :lib:kpipe-format-avro:test :lib:kpipe-format-protobuf:test` — green.

### DIFF
--- a/lib/kpipe-format-avro/src/main/java/io/github/eschizoid/kpipe/format/avro/AvroFormat.java
+++ b/lib/kpipe-format-avro/src/main/java/io/github/eschizoid/kpipe/format/avro/AvroFormat.java
@@ -138,7 +138,10 @@ public final class AvroFormat implements MessageFormat<GenericRecord> {
       encoder.flush();
       return output.toByteArray();
     } catch (final IOException e) {
-      throw new RuntimeException("Failed to serialize Avro record", e);
+      throw new RuntimeException(
+        "AvroFormat.serialize failed for record with schema " + data.getSchema().getFullName(),
+        e
+      );
     }
   }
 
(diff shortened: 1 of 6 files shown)


eschizoid:
@copilot please review

PR #208, opened 2026-06-20
fix: include offending value + remediation in validation error messages

eschizoid:
## What

Validation throws that stated the rule but omitted the offending value (or omitted how to fix). Several sibling validations in the same files already do it right (include `got ` + value); this PR makes them consistent. These are message-only changes on existing throw sites — **exception types are unchanged**.

## Fixes

1. **`BackpressureController` ctor** — `highWatermark must be positive` / `lowWatermark cannot be negative` now append `, got <value>`, matching the existing `lowWatermark (%d) must be less than highWatermark (%d)` sibling.
2. **`KafkaOffsetManager.Builder.withCommitInterval`** — `Commit interval must be positive` → appends `, got <interval>`.
3. **`KPipeConsumer.Builder.build`** — `Poll timeout must be positive` / `Max retries cannot be negative` now append the got-value, matching the sibling `maxKeys`/`interval` checks.
4. **`DefaultStream.withBackpressure(long,long)`** — `Invalid watermarks: high must be positive and > low (got high=%d, low=%d)` → `withBackpressure requires high > low > 0 (got high=%d, low=%d)`. Names the entry point, drops the meta-word "Invalid".
5. **`MultiBuilder.start`** — `at least one route is required` → names the API and the route methods to call before `start()`.
6. **(LOW)** **`KafkaOffsetManager.Builder.build`** — `Command queue must be provided` → `withCommandQueue(...) must be called before build()`. Escape-hatch API, low urgency, cheap.

Out of scope: the `ConsumerHealthController` "hook cannot be null" finding (internal-ish constructor param, not user-facing).

## Tests

- Added focused got-value tests for the two most user-facing cases: `BackpressureControllerTest.shouldIncludeOffendingValueInWatermarkMessages` and `StreamCompositionTest.withBackpressureRejectsInvalidWatermarksWithOffendingValues`.
- No existing test asserted any of the changed strings, so nothing needed updating. The trivial string-only changes are not over-tested.

`./gradlew :lib:kpipe-consumer:test :lib:kpipe-api:test` is green except two Testcontainers integration tests that fail to initialize because no Docker daemon is available in this environment (`DockerClientProviderStrategy` init error) — unrelated to this change.

### DIFF
--- a/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/DefaultStream.java
+++ b/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/DefaultStream.java
@@ -157,7 +157,7 @@ record DefaultStream<T>(
   @Override
   public Stream<T> withBackpressure(final long high, final long low) {
     if (low < 0 || low >= high) throw new IllegalArgumentException(
-      "Invalid watermarks: high must be positive and > low (got high=%d, low=%d)".formatted(high, low)
+      "withBackpressure requires high > low > 0 (got high=%d, low=%d)".formatted(high, low)
     );
     return mutate(m -> {
       m.backpressureHigh = high;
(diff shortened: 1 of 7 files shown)


eschizoid:
@copilot please review

PR #210, opened 2026-06-20
fix(logging): dedup unrouted-key WARN, drop per-record INFO spam, attach throwables

eschizoid:
Logging-hygiene roundup from the logging audit.

## A. Per-record spam (HIGH)
- **Registry unrouted-key WARN deduped.** The per-record WARN for an unrouted operator/sink key now fires **once per distinct key** (`ConcurrentHashMap.newKeySet()` guard, `Set.add()` as the one-shot gate — mirrors the one-shot-WARN pattern elsewhere in the codebase). The WARNING is load-bearing (operators need to know a key is unrouted) so it **stays at WARNING** — only the repeat is suppressed. Separate sets per namespace so the same key still warns once on the operator side and once on the sink side (the two lookups are independent). Bounded by `RegistryKey` cardinality.
- **`KPipeConsumer` retry log INFO -> DEBUG** (retry count already lives in `METRIC_RETRIES`).
- **`KPipeProducer` "Sent record to DLQ topic" INFO -> DEBUG** (`METRIC_DLQ_SENT` already counts).

## B. Missing throwable / dropped stack trace (MEDIUM)
Defensive tracer/span/sink shields and the per-record + batch failure logs that passed only `e.getMessage()` now actually attach the throwable.

> **Audit finding that changed the approach.** The task brief assumed `System.Logger` strips a trailing `Throwable` from positional substitution. That is **not true for the JDK `System.Logger` -> `java.util.logging` bridge** (`sun.util.logging.internal.LoggingProviderImpl$JULWrapper`). I verified empirically: the bridge only routes a trailing arg to `LogRecord.setThrown` when the message has **no `{N}` placeholders**. With any `{N}` present, the trailing throwable is consumed as a format param and — with no matching placeholder — silently dropped (no stack trace). So `"{0}: {1}", offset, e.getMessage(), e` would have logged the message but lost the stack trace.
>
> These sites therefore use either the placeholder-free `"message", throwable` form, or the lazy `Supplier<String> + Throwable` overload (lazy **and** attaches the throwable), instead of positional `{0}`/`{1}` + trailing throwable.

Sites: `OtelTracer` (7 catch blocks), `KPipeConsumer` (`startConsumerSpan` / `SpanScope.close` / `recordException` shields + per-record failure + batch-failure callback), `KPipeProducer` (tracer inject), `KafkaMessageSink` (tracer inject + send callback), `BatchPipelineWrapper` (coverage-violation WARN).

## C. Eager string building at log sites (MEDIUM/LOW)
`.formatted()` / `+` concat at log sites converted to lazy `Supplier<String>` where the throwable must stay attached (KafkaMessageSink, KPipeProducer DLQ-failure, DefaultSink observer, KPipeConsumer rejected-task), or positional `{0}` where no throwable is involved.

## D. Per-record ERROR -> WARNING (HIGH)
`CompositeMessageSink` is explicitly best-effort (continues past a failing sink); a sustained downstream outage at ERROR floods. Dropped to WARNING.

## E. Interrupt visibility (LOW)
`ParallelDispatcher.close()` now logs a WARNING (with the exception) when interrupted while awaiting executor termination, before `shutdownNow()`.

## Tests
- `MessageProcessorRegistryUnroutedKeyWarnTest` — pins the dedup: WARN fires exactly once per distinct key across 50 repeated lookups, and operator/sink namespaces dedup independently. Captured via the JUL bridge (same pattern as the KEY_ORDERED saturation test).
- `BatchPipelineWrapperCoverageContractTest.coverageViolationWarnCarriesThrowable` — asserts the coverage-violation WARN's `LogRecord.getThrown()` is the synthetic `IllegalStateException` (would be null if a regression dropped the cause).

## Test run
`./gradlew :lib:kpipe-consumer:test :lib:kpipe-producer:test :lib:kpipe-core:test :lib:kpipe-tracing-otel:test`

All unit tests pass. The only 2 failures are `ExternalOffsetIntegrationTest` / `KPipeProducerIntegrationTest`, which are `@Testcontainers` integration tests failing at `DockerClientProviderStrategy` because no Docker daemon is available in the build environment — unrelated to these changes; they pass where Docker is present.

## Conventions
No `@Deprecated` (delete + migrate same PR), `java.lang.System.Logger`, `final var`, no FQCN, no spec/section refs in comments — rules stated directly.

### DIFF
--- a/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/DefaultSink.java
+++ b/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/DefaultSink.java
@@ -102,7 +102,7 @@ record DefaultSink<T>(DefaultStream<T> stream, MessageSink<T> terminalSink) impl
     try {
       observer.accept(arg);
     } catch (final RuntimeException e) {
-      LOGGER.log(Level.WARNING, name + " observer threw; swallowing to keep the pipeline running", e);
+      LOGGER.log(Level.WARNING, () -> name + " observer threw; swallowing to keep the pipeline running", e);
     }
   }
 
--- a/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapper.java
+++ b/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/BatchPipelineWrapper.java
@@ -216,7 +216,7 @@ final class BatchPipelineWrapper<K, T> implements AutoCloseable {
           size +
           " — treating as failures to avoid silent data loss"
       );
-      LOGGER.log(Level.WARNING, coverageViolation.getMessage());
+      LOGGER.log(Level.WARNING, coverageViolation.getMessage(), coverageViolation);
     }
 
     for (int i = 0; i < size; i++) {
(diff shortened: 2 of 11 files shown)


eschizoid:
@copilot please review

PR #211, opened 2026-06-20
docs: harden audit-review comments on non-obvious correctness

eschizoid:
## Summary

Four documentation hardenings surfaced as **Worth Considering** during the R1 convergence review of the audit-fix PRs (#205–#210). All comment/javadoc only — zero behavior change, no new tests needed.

Each documents *why* a piece of non-obvious code is correct, so a future maintainer doesn't "fix" it into a bug:

| Source PR | Hardening |
|---|---|
| #207 | `AvroFormat.deserializeStatic` — why the catch is broadened to `RuntimeException` (Avro throws `AvroRuntimeException` for malformed bytes, not `IOException`) + the cause is always attached so real bugs stay visible |
| #209 | `KPipeConsumer.closePartiallyConstructed` — why reading the `final` fields compiles (separate method ⇒ definite-assignment doesn't guard the read ⇒ unreached blank-final reads `null`); explicit "don't drop `final` to fix the apparent unassigned read" |
| #210 | `MessageProcessorRegistry` — the one unbounded case for the unrouted-key dedup set (dynamically-generated keys) + remedy if it ever becomes real |
| #205 | `OffsetManagerBoxingBenchmark` — drop the JVM-specific object-header byte estimate; point at the measured `gc.alloc.rate.norm` delta (prove, don't assert) |

## Test plan

- [x] `./gradlew :lib:kpipe-format-avro:test :lib:kpipe-core:compileJava :lib:kpipe-consumer:compileJava :benchmarks:compileJmhJava` → BUILD SUCCESSFUL
- [ ] Merge to main

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/kpipe/benchmarks/OffsetManagerBoxingBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/kpipe/benchmarks/OffsetManagerBoxingBenchmark.java
@@ -58,10 +58,11 @@ import org.openjdk.jmh.annotations.Warmup;
 /// ```
 ///
 /// The `gc.alloc.rate.norm` column (bytes allocated per op) is the primary signal — the cache
-/// should remove one `TopicPartition` instance worth of bytes per record (16 bytes object header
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #212, opened 2026-06-20
ci: add Benchmarks workflow (dispatch + weekly bench-rot smoke)

eschizoid:
## Summary

Adds `.github/workflows/benchmarks.yaml` to run the JMH suite — deliberately **not** on PRs (slow + noisy-runner numbers are unreliable), but in two useful modes:

### `workflow_dispatch` — run on demand
Inputs: `includes` (JMH class regex), `iterations`, `warmup`, `fork`, `profilers`. Defaults to the five Docker-free benches. Use it for branch-vs-main deltas under pinned conditions.

### `schedule` — weekly bench-rot smoke (Mondays 06:00 UTC)
Low-iteration run of the Docker-free benches. Its real value is **catching bench-rot automatically**: `failOnError = true` in `benchmarks/build.gradle.kts` means a benchmark that no longer compiles or throws at runtime fails this job. That's exactly the regression class that silently broke `JsonPipelineBenchmark` through an architecture change and forced its restoration in #154 — now it surfaces within a week instead of whenever someone next runs the suite.

## Output
- `results.json` uploaded as an artifact (90-day retention)
- A throughput digest table written to the job summary, with an explicit caveat that shared GitHub runners make these smoke / coarse-regression signal only — **not** absolute numbers

## Benchmark classification (why the smoke set is what it is)
| Bench | Harness | In weekly smoke? |
| --- | --- | --- |
| JsonPipelineBenchmark | in-process `MessagePipeline.process` | ✅ |
| AvroPipelineBenchmark | in-process `MessagePipeline.process` | ✅ |
| OffsetManagerBoxingBenchmark | synthetic records | ✅ |
| KeyOrderedDispatchBenchmark | MockConsumer | ✅ |
| BatchSinkLatencyBenchmark | MockConsumer | ✅ |
| ParallelProcessingBenchmark | Testcontainers Kafka | dispatch-only (heavier) |
| ParallelProcessingLatencyBenchmark | Testcontainers Kafka | dispatch-only (heavier) |

## Test plan
- [x] YAML validated
- [x] Smoke-ran `:benchmarks:jmh -Pjmh.includes=OffsetManagerBoxing -Pjmh.iterations=1` locally → `results.json` produced at the expected path
- [x] Summary `jq` expression verified against real output, including the `scoreError == "NaN"` (too-few-iterations) edge case → renders `n/a`
- [ ] After merge: trigger a manual `workflow_dispatch` run to confirm the Actions environment end-to-end

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,139 @@
+name: Benchmarks
+
+# JMH benchmarks are NOT run on every PR — they are slow, and GitHub-hosted runners are shared,
+# noisy CPU environments where absolute throughput numbers are unreliable. This workflow exists
+# for two things:
+#   1. workflow_dispatch — run a chosen benchmark set on demand with tunable rigor (iterations,
+#      warmup, fork). Use this when you want numbers to compare a branch against main; pin the
+#      runner conditions and read the delta, not the absolute value.
+#   2. schedule (weekly) — a low-iteration smoke of the Docker-free benches. Its real job is
+#      catching bench-rot: `failOnError = true` in benchmarks/build.gradle.kts means a benchmark
+#      that no longer compiles or throws at runtime fails this job. (The JsonPipelineBenchmark
+#      silently rotted through an architecture change once and had to be restored; this catches
+#      that class of regression automatically.)
+
+on:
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #215, opened 2026-06-21
bench: annotation-driven benchmark mode (restores CI latency percentiles) + CI regime docs

eschizoid:
## Summary

Two bench-infra follow-ups from the 2026-06-20 CI capture, bundled.

### #1 — Restore latency percentiles in CI
`benchmarks/build.gradle.kts` set `benchmarkMode = listOf("thrpt")` globally, which the JMH plugin applies over every benchmark — overriding `ParallelProcessingLatencyBenchmark`'s own `@BenchmarkMode({SampleTime, AverageTime})`. That's why the CI capture produced a second throughput dataset instead of p50/p95/p99.

Fix: **delete the global `benchmarkMode`/`timeUnit`** and let each class's annotation be the source of truth. Verified every bench is already annotated:
- throughput benches → `@BenchmarkMode(Mode.Throughput)` + `SECONDS` (no behavior change)
- latency bench → `@BenchmarkMode({SampleTime, AverageTime})` + `MILLISECONDS` (now honored → percentiles)

### #2 — Document the CI-runnable regimes
`METHODOLOGY.md` now records what fits CI now that `-Pjmh.workMicros` exists:
- **low regime** `0,100,1000` — fits easily
- **high regime** `10000,35000,50000,100000` — fits **only with the serial arms excluded** (`singleThread` is ~42 min/iter at 100ms; `confluentPartition` is 8-way capped), via a `$`-anchored include regex so `confluent` doesn't also match `confluentPartition`
- **latency percentiles** — captured by just running the latency bench (annotation drives mode)
- the **full sweep incl. `singleThread`** stays a local overnight run

## Verification
- [x] `:benchmarks:jmhJar` config parses with the globals removed
- [x] confirmed all 7 bench classes carry their own `@BenchmarkMode`
- [ ] post-merge: dispatch the latency bench in CI to confirm percentiles emit (Testcontainers — can't verify locally with Docker down)

## Note
Does not re-capture numbers — purely the mode fix + docs. The percentile + high-regime captures are follow-on dispatches once this is on main.


eschizoid:
@copilot please review

PR #216, opened 2026-06-21
bench: record allocation attribution (profile-first finding for #3)

eschizoid:
## Summary

Banks the profile-first attribution of KPipe PARALLEL's 1,628 B/op (flagged in `2026-06-20.md`) as design memory.

`gc.alloc.rate.norm` is deterministic, so it's trustworthy even though this laptop's throughput numbers weren't. Measured two of three hot-path components in isolation via the Docker-free benches.

| Path | B/op | Share |
|---|---:|---:|
| Dispatcher + VT creation | ~1,010 | ~62% |
| Offset track+mark (Long boxing) | 396 | ~24% |
| Pipeline (deserialize + Result + ops) | ~220 | ~14% |

**Conclusions:** `Result.Passed` is noise (don't chase it); VT creation is the price of the model (the same thing that wins under blocking work — documented, not cut); the only addressable lever is the offset manager's 396 B/op, **deferred into the correctness phase** because it lives in the concurrency-critical lowest-pending-offset structure (optimize under a Lincheck harness, not before).

Docs only.

### DIFF
--- a/benchmarks/results/2026-06-21-allocation-attribution.md
+++ b/benchmarks/results/2026-06-21-allocation-attribution.md
@@ -1,0 +1,56 @@
+# Allocation attribution — 2026-06-21
+
+Follow-up to [`2026-06-20.md`](2026-06-20.md), which flagged KPipe (PARALLEL) at **1,628 B/op** — the highest per-record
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #217, opened 2026-06-21
test(consumer): correctness suite — offset invariants (jcstress + jqwik + stress), chaos rebalance, circuit-breaker & watermark

eschizoid:
## At-least-once correctness suite + 3 production fixes

This PR hardens and *verifies* kpipe's at-least-once guarantee, end to end. The verification is real, not green-for-green's-sake — it found and fixed three genuine bugs.

### Production bugs found & fixed
1. **`trackOffset` pending-set data loss** (found by jcstress). `pendingOffsets.computeIfAbsent(p, …).add(offset)` ran the `.add()` outside the bucket lock; it raced `markOffsetProcessed`'s remove-if-empty and could land on a set already orphaned out of the map → a pending offset silently lost → commit point advances past an in-flight record. Fixed by doing the add inside `compute(…)`. Pinned by `RemoveIfEmptyJCStressTest`.
2. **DLQ-send-failure silent loss.** `handleProcessingError` / `onBatchFailure` marked the offset processed even when `sendToDlq` returned `false` → record neither processed, parked, nor reprocessed. Now the offset is marked only on a durable terminal state (sink success or successful DLQ park); a failed send holds the offset (reprocessed on restart) and increments a new `dlqFailed` / `kpipe.producer.dlq.failed` counter. Pinned by `DlqSendFailureTest` (per-record + batch).
3. **Circuit breaker blind to the batch path.** Batch outcomes never fed `health.recordOutcome`, so a failing batch sink could never trip the breaker (and could trip spuriously). Now both batch outcomes feed the window symmetrically with the per-record path.

### ⚠️ Operator-facing behavior change
A **persistently-down DLQ now stalls a partition's commit point** at the first undeliverable record instead of marking-and-advancing (the old silent-drop). This is the honest at-least-once tradeoff: it's a **commit stall, not backpressure** — the in-memory fetch position keeps advancing, the record is reprocessed only on restart/rebalance, and a permanently-undeliverable record (oversized/serializer/authz) pins the commit indefinitely. Observable via the `dlqFailed` counter. Documented in `docs/OFFSET-INVARIANTS.md` (I3).

### Verification added
- **jcstress — 15 interleaving tests across 4 modules** (consumer, core, schema-registry, producer), now **run as a gating step in CI** (capped depth; ~100k+ interleavings/test, fails on any FORBIDDEN outcome). Covers: offset lowest-pending / gap-hold / revoke-race / remove-if-empty / safe-first, state-machine CAS, CAS-publication, parallel-dispatcher pending-count, key-ordered queue handoff, circuit-breaker window, batch-wrapper lock, registry namespaces, schema-cache miss-collapse, producer metrics, composite fanout.
- **jqwik** property tests + **virtual-thread stress** tests for the offset lifecycle (I1–I5).
- **Testcontainers E2E:** `ChaosRebalanceIntegrationTest` (rebalance handoff) + `CrashRestartReprocessingIntegrationTest` (hard crash + restart). Both hard-assert **no-loss** and **no-commit-ahead**.

### Honest scoping
- **`CrashRestartReprocessingIntegrationTest` re-delivery overlap is logged, not asserted.** In a single-broker container the A→B overlap is timing-dependent (A may commit its full processed set right before the crash — still no loss). The hard gates remain no-loss + no-commit-ahead; the "uncommitted offset is reprocessed" property is proven deterministically by `RemoveIfEmptyJCStressTest`, the offset property/stress suites, and `DlqSendFailureTest`.
- "Verified by multiple independent methods," not "proven correct everywhere."

### Deferred (follow-ups, non-blocking)
- `CircuitBreakerWindowJCStressTest` and `StateTransitionCasJCStressTest` mirror the real (private) logic rather than driving it — disclosed in their javadocs; a refactor to expose the real `SlidingWindow` could remove the replica-drift risk.
- Transient-vs-permanent DLQ classification (a permanently-undeliverable record wedges its partition); a periodic stall-heartbeat log.
- `CLAUDE.md` §10/§11 sync (local instructions; gitignored).


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review — added a fix for a DLQ silent-loss bug (offset was marked processed even when the dead-letter send failed) plus correctness tests for the retry-interrupt, DLQ-failure, batch coverage-contract, shutdown-ordering, and multi-topic-unrouted paths.

eschizoid:
@copilot please review — this branch now adds a full at-least-once correctness suite (jcstress interleavings across 5 modules, jqwik properties, VT stress, chaos rebalance + crash-restart E2E) plus three production fixes: the trackOffset pending-set atomicity (jcstress-found data loss), the DLQ-send-failure offset hold, and feeding the circuit breaker on the batch path.

PR #220, opened 2026-06-22
test(consumer): Tier-1 correctness hardening — real-class jcstress + deterministic crash-restart

eschizoid:
## Tier-1 correctness hardening (follow-up to #217)

Closes two of the gaps #217 disclosed. (The third — transient-vs-permanent DLQ-send classification — was deliberately **not** done: existing `dlqFailed` + ERROR-log alerting plus the retry policy are sufficient; keep the strict never-lose hold.)

### 1. Drive the two replica jcstress tests against the real code
The `CircuitBreakerWindow` and `StateTransitionCas` jcstress tests previously reimplemented the production logic in the test (replica-drift risk: a change to the real code wouldn't be caught). Now they drive the real thing:
- `ConsumerHealthController.SlidingWindow`: `private` → **package-private**, so `CircuitBreakerWindowJCStressTest` exercises the real window (replica deleted) — same approach as the existing jcstress tests that reach `KeyOrderedDispatcher`/`KafkaOffsetManager`.
- Extracted `KPipeConsumer.tryTransitionToClosing(AtomicReference<ConsumerState>)` — the static single-read-CAS the instance `transitionToClosing()` now delegates to — so `StateTransitionCasJCStressTest` drives the real transition (replica deleted).

No behavior change; both verified green locally (`:lib:kpipe-consumer:jcstress`, 408/408).

### 2. Deterministic crash-restart re-delivery (hard assertion restored)
#217 had to downgrade `CrashRestartReprocessingIntegrationTest`'s re-delivery overlap to a *logged* best-effort because the 1s auto-commit raced the crash and committed the whole stream, leaving no tail (flaky 4/4 in CI). New **manual-commit harness** removes the timing dependency:
- A's `KafkaOffsetManager` uses a 1-hour auto-commit interval — the only commit is a single manual `commitSyncAndWait` of a small prefix.
- The test then lets A process a known tail (`committed + TAIL_MARGIN`) with no further commit, captures A's observed set, and crashes (stop + interrupt, no graceful drain).
- Everything past the prefix is deterministically uncommitted, so B re-fetches it on restart — the intersection assertion now holds every run.

No-loss and no-commit-ahead remain hard gates. **CI-run-required** (Testcontainers; can't run locally without Docker — the determinism is verified by this CI run).

### Verification
- jcstress (4 modules, CI-gated since #217) green locally: 408/408, 0 forbidden.
- Crash-restart compiles; its determinism is proven by this PR's CI run.

### DIFF
--- a/lib/kpipe-consumer/src/jcstress/java/io/github/eschizoid/kpipe/consumer/CircuitBreakerWindowJCStressTest.java
+++ b/lib/kpipe-consumer/src/jcstress/java/io/github/eschizoid/kpipe/consumer/CircuitBreakerWindowJCStressTest.java
@@ -1,7 +1,5 @@
 package io.github.eschizoid.kpipe.consumer;
 
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicLongArray;
 import org.openjdk.jcstress.annotations.Actor;
 import org.openjdk.jcstress.annotations.Arbiter;
 import org.openjdk.jcstress.annotations.Expect;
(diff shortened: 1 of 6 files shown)


eschizoid:
@copilot please review — Tier-1 follow-up to #217: drives the cb-window + state-CAS jcstress tests against the real classes (removing replicas), and restores a deterministic, hard-asserted crash-restart re-delivery test via a manual-commit harness.

PR #228, opened 2026-07-10
fix(consumer): lost-wakeup deadlock when backpressure pause races a fast in-flight drain

eschizoid:
## What

Diagnosed as the 'PARALLEL × low-regime single-sample' benchmark anomaly — root cause is a **production-reachable liveness bug**, not a harness artifact.

**The race** (`ConsumerHealthController.tickBackpressure` + `KPipeConsumer.afterRecordComplete`): the pause decision reads the in-flight metric, then logs + fires hooks, and only *then* publishes the BACKPRESSURE bit. Completions only unpark the consumer if they observe that bit after decrementing. At fast per-record work (~100µs), 10k in-flight records drain in milliseconds — inside that window — so every completion misses the bit, the consumer parks with zero records in flight, and nothing ever wakes it. CI logs show the signature: pause at in-flight=10,000 → 2 minutes of silence → timeout, with processed counts always a multiple of max.poll.records.

**Why only PARALLEL × fast cells**: that combination drains the whole in-flight population inside the publish window almost every time; KEY_ORDERED drains slower through per-key queues (still flaked once in the 2026-06-20 capture — same race, narrower window); at ≥10ms work the drain always straddles the bit-set.

**Fix**: after publishing the pause bit, re-check the watermark — if the metric already fell to/below the low watermark, release immediately; otherwise at least one in-flight record is guaranteed to see the bit and unpark (Dekker-style handshake via volatile total order; full reasoning in the code comment). RESUME choreography extracted into a shared `releaseBackpressure()`. Deterministic regression test included that reproduces the deadlock without the guard.

## Validation
- 376 kpipe-consumer unit tests pass locally (4 Docker-required integration tests skip as expected)
- New regression test `tickBackpressureResumesWhenInFlightDrainsInsidePauseWindow`
- CI confirmation benchmark run dispatched on this branch (PARALLEL arm, fork=5, wM=100,1000) — success = 25 rawData samples per cell, no NaN
- Fixes the last data blemish: PARALLEL low-regime cells become fully sampled

### DIFF
--- a/lib/kpipe-consumer/src/jcstress/java/io/github/eschizoid/kpipe/consumer/BackpressureHandshakeJCStressTest.java
+++ b/lib/kpipe-consumer/src/jcstress/java/io/github/eschizoid/kpipe/consumer/BackpressureHandshakeJCStressTest.java
@@ -1,0 +1,92 @@
+package io.github.eschizoid.kpipe.consumer;
+
+import java.util.concurrent.atomic.AtomicLong;
+import org.openjdk.jcstress.annotations.Actor;
+import org.openjdk.jcstress.annotations.Expect;
+import org.openjdk.jcstress.annotations.JCStressTest;
+import org.openjdk.jcstress.annotations.Outcome;
+import org.openjdk.jcstress.annotations.State;
+import org.openjdk.jcstress.infra.results.II_Result;
+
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #229, opened 2026-07-10
perf(consumer): primitive-long pending-offset set — −45% offset-manager allocation

eschizoid:
## What

Replaces the per-partition `ConcurrentSkipListSet<Long>` in `KafkaOffsetManager` with `PendingOffsetSet` — a package-private, monitor-synchronized sorted primitive-`long` window. Closes the 396 B/op lever identified in [`benchmarks/results/2026-06-21-allocation-attribution.md`](benchmarks/results/2026-06-21-allocation-attribution.md), which was deliberately deferred until the correctness harness existed (#217/#220/#221 — it now does, and every suite runs against the new structure).

**Design:** inline single-element mode (evict-on-empty recreates sets constantly — keeps that churn to one small object) spilling to a `long[]` window with two-sided slack; O(1) fast paths for Kafka's dominant orders (monotonic append, head removal); out-of-order ops binary-search + arraycopy; zero steady-state allocation. The §14 bucket-locked `compute`/`computeIfPresent` remove-if-empty patterns are preserved verbatim; `firstOrNull` reads under one monitor acquisition, so the skiplist `first()`-throws race is structurally impossible (`safeFirst`/`safeLast` deleted). No public API changes.

## Measured (JMH `gc.alloc.rate.norm`, deterministic)

| | B/op |
|---|---|
| Baseline (skiplist) | 419.7 |
| **This PR** | **232.0 (−44.7%)** |

Residual 232 is the mandated evict/recreate cycle + lambda captures, not the structure — high-in-flight regimes allocate nothing, so the production win skews larger.

## Verification

- **jcstress: 510/510 pass** — the 5 pre-existing offset suites now exercise the new structure, plus 3 NEW suites (`PendingOffsetSet{AddRemove,ConcurrentAdd,FirstRead}JCStressTest`) covering add-vs-drain, concurrent out-of-order adds, first-read-vs-drain.
- `:lib:kpipe-consumer:test` 390 tests green locally (4 Testcontainers classes need Docker — CI covers them).
- New `PendingOffsetSetTest`: deterministic spill/compaction/negative-offset edges + 500-try jqwik property against a `TreeSet<Long>` oracle after every op.
- Docs updated: OFFSET-INVARIANTS vocabulary, README, SafeFirst jcstress narrative.

## Residual risks (documented in class Javadoc)
Spilled sets retain their high-water array until fully drained (bounded per partition); pathological fully-random completion over a very deep window shifts cost from allocation to arraycopy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,8 @@ KPipe runs on Java virtual threads ([Project Loom](https://openjdk.org/projects/
 KPipe never commits past an in-flight record. The implementation:
 
 - `OffsetManager` is an interface — Kafka-backed by default, but you can plug in external storage.
-- Every in-flight offset is tracked in a `ConcurrentSkipListSet` per partition (see `KafkaOffsetManager`).
+- Every in-flight offset is tracked in a sorted primitive-`long` pending set per partition (see `KafkaOffsetManager` /
+  `PendingOffsetSet` — allocation-free on the track/mark hot path).
 - Offset 102 cannot be committed until 101 finishes, even if 102 completes first. No gaps.
 - On crash, the consumer resumes from the last committed offset. Some records may be reprocessed — standard
   at-least-once behavior — but nothing is skipped.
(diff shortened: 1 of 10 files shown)


eschizoid:
@copilot please review

PR #230, opened 2026-07-10
bench: per-record latency harness — true tail latency (p50/p95/p99/max)

eschizoid:
## What

Closes the 'true per-record latency bench' follow-up from [`benchmarks/results/2026-07-09.md`](benchmarks/results/2026-07-09.md): the existing `ParallelProcessingLatencyBenchmark`'s `@OperationsPerInvocation` division makes its percentiles **batch-amortized**, not tail latency — its numbers were explicitly ruled non-quotable as p99.

**Design** (plain `main()` harness + `:benchmarks:perRecordLatency` JavaExec task — JMH SampleTime can't ingest externally-measured timestamp pairs):
- Producer stamps each payload with its **intended** send slot (`nanoTime`), not the actual send — coordinated-omission correction, so producer stalls surface as latency instead of hiding
- Sink computes `now − stamp` into a preallocated `long[]`; single sort → exact nearest-rank p50/p95/p99/max, no binning error
- 4 arms (KPipe PARALLEL, KPipe KEY_ORDERED, CPC UNORDERED, raw+VT), fresh topic + group per arm, shared broker; warmup excluded via a phase byte; paced at 2,000 rec/s (below all arms' saturation — above it you measure backlog depth, not latency)
- **Clock caveat documented**: producer+consumer share one JVM by design (`nanoTime` is per-process); values compare arms against each other, not production SLAs

## Run

`./gradlew :benchmarks:perRecordLatency` (Docker required; `-Platency.*` flags for workMicros/rate/records/arms). Output: console table + JSON.

## Validation

Compiles (`:benchmarks:compileJmhJava`); task graph verified via `--dry-run`. **Not yet runtime-validated** — no Docker locally and the benchmarks workflow only dispatches `:benchmarks:jmh`; the first real capture is a follow-up (either a workflow tweak or a local run). METHODOLOGY.md updated with the new section + a correction to the sample-time reading guidance. Only library-adjacent change: `EmbeddedKafkaBackend` private → package-private.

### DIFF
--- a/benchmarks/METHODOLOGY.md
+++ b/benchmarks/METHODOLOGY.md
@@ -221,6 +221,8 @@ Reading the numbers:
 - **The measured quantity is intended-produce → work-complete**: producer send (`linger.ms=0`), broker hop over Docker
   loopback, consumer poll, framework dispatch, and the simulated work. Compare arms against each other; the absolute
   values are not production SLA numbers.
+- **Arms run in declaration order against one shared broker**, so OS page cache and broker-JVM warmth accumulate — the
+  last arm sees a warmer system than the first. For order-sensitivity, shuffle `-Platency.arms` across runs.
 - **Keep the rate below every arm's saturation throughput** (defaults: 2,000 rec/s at 100 µs work — far below all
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #233, opened 2026-07-11
fix(consumer): paused consumer keeps polling — lag-pause park-forever + max.poll.interval eviction

eschizoid:
## What

Fixes the CRITICAL pre-existing liveness bug found by #228's adversarial review (finding A1): with the lag strategy (auto-derived in SEQUENTIAL mode), a backpressure pause parked the consumer thread with **no reachable wake source** — SEQUENTIAL has no async completions, and lag is monotonically non-decreasing while parked (frozen position vs growing endOffsets). One WARNING, then a permanently wedged consumer; and since a parked consumer stops calling `poll()`, it silently left the group after `max.poll.interval.ms` — a hazard shared by long manual and circuit-breaker pauses.

## Design: paused = keep polling (not bounded park)

The paused branch now flushes the command queue, defensively re-issues `pause(assignment())` (per-partition pause doesn't survive revoke/assign; also closes a state-flipped-before-command race), then falls through to the normal poll. Paused partitions fetch nothing, but:
- **group membership stays alive** — Kafka's documented pause+poll contract, fixing the eviction hazard for ALL pause sources (a bounded-park design would not)
- each iteration is bounded by `pollTimeout`, so `tickBackpressure` re-evaluates lag on a fixed cadence and resumes on external lag drops (reassignment, retention truncation, offset reset)
- records slipping through a mid-poll rebalance are processed, never dropped (positions already advanced — dropping would lose data on commit)
- single-writer discipline preserved; no public API changes

Documented tradeoff: resume latency is now ≤ `pollTimeout` (100ms default) instead of an instant unpark.

## Verification
- New `PausedPollLivenessTest` (3 tests, bounded awaits, MockConsumer): **all 3 verified failing against the old loop**, passing with the fix
- `:lib:kpipe-consumer:test`: 397/0/0 · jcstress: 16/16
- Residual (documented): sustained-high-lag pause still doesn't self-resume — inherent to lag-watermark semantics; the consumer is now live, in-group, resumable, and observable instead of wedged

### DIFF
--- a/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/ConsumerHealthController.java
+++ b/lib/kpipe-consumer/src/main/java/io/github/eschizoid/kpipe/consumer/ConsumerHealthController.java
@@ -193,8 +193,9 @@ final class ConsumerHealthController {
         // Every record that completed inside that window skipped the unpark — and when the
         // count drains all the way down in there (routine for PARALLEL mode with fast records:
         // 10k in-flight at ~100µs each finish in a few milliseconds, faster than the log + hook
-        // calls above), no completion ever arrives to wake the consumer, which then parks
-        // forever with nothing in flight. Re-checking with the bit visible closes the window:
+        // calls above), no completion ever arrives to nudge the consumer, which then sits paused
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #234, opened 2026-07-11
test(consumer): de-flake CB batch test — PARALLEL exposed a real pause-delivery race

eschizoid:
## The flake was a product bug

`sustainedBatchFailuresTripBreaker` (5 CI failures in 2 days, blocking unrelated PRs) was the ONLY test in the class running in default PARALLEL mode — so the breaker tripped on a batch-flush worker thread, exposing a real cross-thread race: `internalPause()` CAS-flips state to PAUSED **before** offering the Pause command. The consumer thread can observe `isPaused()`, drain a still-empty command queue, and `park()` without ever executing `kafkaConsumer.pause()` — and a CB pause has no unpark source until the half-open probe (30s here), so the 5s await expires.

**Evidence:** 400-repetition harness → **363/400** reproductions pre-fix (thread dump shows the consumer VT parked at the `isPaused()` branch with the Pause command undelivered); **0/400** in SEQUENTIAL; 4 consecutive `--rerun-tasks` suite runs green.

## Fix here vs fix of the bug

This PR: one line — `.withProcessingMode(ProcessingMode.SEQUENTIAL)`, matching every sibling test. Mode-orthogonal claim ("batch failures feed the breaker window") still fully asserted; the 400-rep harness also drops from 12m19s to 15s.

The **underlying product bug** (state-flipped-before-command-queued) is fixed by open **PR #233**, which retires the park path and defensively re-issues `pause(assignment())` — it names this exact race. This test fix is complementary, touches nothing #233 touches, and unblocks CI immediately.

### DIFF
--- a/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KPipeCircuitBreakerIntegrationTest.java
+++ b/lib/kpipe-consumer/src/test/java/io/github/eschizoid/kpipe/consumer/KPipeCircuitBreakerIntegrationTest.java
@@ -87,11 +87,23 @@ class KPipeCircuitBreakerIntegrationTest {
     // per-record path — otherwise a failing batch sink could never trip the breaker. ofSize(1)
     // flushes each record inline, so every record drives one failure outcome into the window;
     // windowSize=5 + threshold=0.5 trips once the window fills with all-failures.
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #206, opened 2026-06-20
chore(jpms): remove unused requires, promote API-exposed deps to transitive

eschizoid:
JPMS audit follow-up: tighten `module-info.java` across the module graph. **Build-metadata only — no behavior change.** Module-graph changes affect downstream consumers, so each one was verified by grep + a full clean compile (`compileJava` + `compileTestJava` across all modules, re-run with `--rerun-tasks` to defeat caching). All passed.

## Part A — remove unused `requires`

| Module | Removed | Verification grep |
|---|---|---|
| `kpipe-api` | `requires kafka.clients` | `grep -rn "org.apache.kafka" lib/kpipe-api/src/main/java/` → **no matches** (exit 1) |
| `kpipe-consumer` | `requires java.net.http` | `grep -rn "java.net.http" lib/kpipe-consumer/src/main/java/` → only match was the `module-info.java` declaration itself; the health server uses `jdk.httpserver` / `com.sun.net.httpserver`, which stays declared |

Both removals confirmed safe: no audit-missed usage, and the clean compile across all modules still succeeds.

## Part B — promote `requires` → `requires transitive`

Each promotion was confirmed by locating the required module's type on a public/exported signature before promoting:

| Module | Promoted | Public-API exposure (verified) |
|---|---|---|
| `kpipe-api` | `org.apache.avro` | `KPipe.avro(...)` returns `Stream<GenericRecord>` |
| `kpipe-api` | `com.google.protobuf` | `KPipe.protobuf(...)` returns `Stream<Message>` |
| `kpipe-consumer` | `kafka.clients` | `Builder.withConsumer(Supplier<Consumer<K,byte[]>>)`, `withKafkaProducer(Producer<K,byte[]>)`, `withOffsetManagerProvider(Function<Consumer<K,byte[]>, …>)` |
| `kpipe-producer` | `kafka.clients` | `KPipeProducer.Builder.withProducer(Producer<K,V>)`, `send(ProducerRecord)`, `sendAsync(ProducerRecord)` |
| `kpipe-metrics-otel` | `io.github.eschizoid.kpipe.metrics` | `OtelConsumerMetrics implements ConsumerMetrics` (and `OtelProducerMetrics implements ProducerMetrics`) |
| `kpipe-tracing-otel` | `io.github.eschizoid.kpipe.producer` | `OtelTracer implements Tracer` (`producer.tracing.Tracer`) |
[...]

PR #238, opened 2026-07-13
chore: clear the fixable -Xlint compiler warnings (89 → 68)

eschizoid:
## What

The build never enabled `-Xlint`, so 89 compiler warnings sat unsurfaced. Ran `-Xlint:all` across every module, triaged, and fixed every actionable one — leaving only genuine noise.

| Category | Before | After | Action |
|---|---|---|---|
| `exports` | 8 | **0** | `kpipe-metrics-otel` + `kpipe-tracing-otel` expose `OpenTelemetry` (an api-module type) in public API → promoted `requires` to `requires transitive` (the §9 / #206 pattern) |
| `deprecation` | 12 | **0** | deprecated `MockConsumer(OffsetResetStrategy)` → String constructor in 3 test files (matches newer tests); unused imports removed |
| `try` | 1 | **0** | intentional never-referenced OTel `scope` → `@SuppressWarnings("try")` + why-comment |
| `unchecked` | 42 | 42 | **left** — all test-code Mockito `mock(Generic.class)` casts, unavoidable |
| `varargs` | 1 | 1 | **left** — `DefaultStream.toMulti` is already `@SafeVarargs`; residual note on the inner `List.of` is a javac quirk |

## Verification
`-Xlint:all` recompile confirms exports/deprecation/try all at 0; the touched modules' test suites pass. The two `requires transitive` changes only *add* readability downstream (can't break callers).

## Note
This is the **compiler-verifiable** subset. IntelliJ's default-profile style inspections (redundant cast, unused declaration, can-be-simplified) are separate — the IDE MCP is disconnected in this environment and the headless inspector crashes, so those aren't covered here.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,7 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="JavaCollectionWithNullableTypeArgument" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="UsePropertyAccessSyntax" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
+  </profile>
+</component>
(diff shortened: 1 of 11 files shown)


eschizoid:
@copilot please review

PR #244, opened 2026-07-16
refactor(formats): standardize read-path failures on IllegalStateException

eschizoid:
Delivers the demoted architecture-critique item: unify the "malformed wire envelope / decode failed" exception type across the format modules.

## The divergence
The same class of failure — bad input bytes at the read boundary — threw three different types:
- **JSON / Avro** deserialize + wire-envelope → `RuntimeException`
- **Protobuf** envelope → `IllegalStateException`, but static/registry **decode** → `RuntimeException`
- **Confluent descriptor compiler** wire-derived message-index errors → `IllegalArgumentException`

A reader grepping logs or catching by type got no consistent signal.

## The standard
Every read-path failure (deserialize, wire-envelope parse, descriptor-index resolution) now throws **`IllegalStateException`** — the type `MessagePipeline.deserializeOrFail` already uses ("null deserialization → IllegalStateException, retryable") and the one Protobuf's envelope path already followed.

- `JsonFormat.deserialize`, `AvroFormat` static-decode + all four envelope conditions, `ProtobufFormat` static/envelope decode → ISE.
- `ConfluentProtobufDescriptorCompiler`'s three message-index errors → ISE.

## Deliberate boundaries (preserved)
- **Resolver-thrown exceptions propagate unchanged** — a transiently-down `SchemaResolver` is not a malformed-envelope condition to reclassify (pinned by `aThrowingResolveIsNotCachedSoTheNextRecordRetries`, which the test-tightening below actually surfaced).
- **Write-path `serialize` failures stay `RuntimeException`** — a can't-encode-this-object failure is a distinct mode from bad-input-bytes.

## Not a correctness change
No consumer path switches on exception type (retry catches `Exception` uniformly), so runtime behavior is unchanged. This is log/reader consistency. `RuntimeException` is a supertype of `IllegalStateException`, so pipeline-level roundtrip assertions still pass unchanged.

## Tests
- The Confluent index test moves `IllegalArgumentException` → `IllegalStateException`.
- The format-unit deserialize-failure assertions are tightened from the loose `RuntimeException` supertype to `IllegalStateException` to **pin** the standard — this tightening is what caught the resolver-propagation boundary above.

## Verification
`./gradlew :lib:kpipe-format-json:test :lib:kpipe-format-avro:test :lib:kpipe-format-protobuf:test :lib:kpipe-format-protobuf-confluent:test` — BUILD SUCCESSFUL.

### DIFF
--- a/lib/kpipe-format-avro/src/main/java/io/github/eschizoid/kpipe/format/avro/AvroFormat.java
+++ b/lib/kpipe-format-avro/src/main/java/io/github/eschizoid/kpipe/format/avro/AvroFormat.java
@@ -159,8 +159,8 @@ public final class AvroFormat implements MessageFormat<GenericRecord> {
       // RuntimeException is caught alongside IOException because malformed bytes surface as Avro's
       // own AvroRuntimeException (e.g. "Malformed data. Length is negative"), not IOException — so
       // the common bad-payload case would otherwise escape with a bare, context-free message. The
-      // original cause is always attached, so a genuine programming error (NPE, IllegalState) is
-      // still visible in the cause chain rather than swallowed.
+      // original cause is always attached, so a genuine programming error (e.g. an NPE) is still
+      // visible in the cause chain rather than swallowed.
       throw new IllegalStateException(
         "AvroFormat.deserialize failed in static mode on " + data.length + " bytes against schema " +
           staticSchema.getFullName(),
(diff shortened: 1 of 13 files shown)


eschizoid:
**Design trade-off surfaced in review (acknowledged, no change):** `IllegalStateException` is now overloaded in the format layer — it signals both genuine state violations (`consoleSink()` in registry mode, no compiler on the runtime path) and malformed input bytes, disambiguated only by the message string. A dedicated `MalformedEnvelopeException` was considered and deliberately rejected: nothing downstream branches on exception type (the consumer catches `Exception` uniformly and treats all as retryable), so a named type would be speculative surface against the §16 no-deprecation ethos. Revisit only if a caller ever needs to route malformed-input straight to DLQ without retry. Write-path `serialize` is left as `RuntimeException` on purpose (a can't-encode failure is genuinely less classifiable than bad input bytes).

PR #245, opened 2026-07-16
refactor(formats): share byte previews via core WireDiagnostics + fix Avro envelope blind spot

eschizoid:
A "derive, don't duplicate" deepening that also closes a real diagnostic gap.

## Problem
`JsonFormat` and `ProtobufFormat` each carried a **byte-identical** private `hexPreview` (same body, `HEX_PREVIEW_LIMIT=8`) — duplicated diagnostic code drifting toward divergence — while `AvroFormat` had **none**. So the single hardest-to-debug path in the library (Avro SR writer/reader mismatch on real bytes) shipped with the least diagnostic context of any format, compounded by a too-narrow catch.

## One move fixes dup + gap
- **New `io.github.eschizoid.kpipe.diagnostics.Diagnostics.hexPreview`** in `kpipe-core` (exported package), shared by every format. JSON and Protobuf delegate to it; their local copies and `java.util.HexFormat` imports are deleted.
- **AvroFormat envelope blind spot:**
  - Widen the envelope-decode `catch(IOException)` → `catch(IOException | RuntimeException)` to match the static path. A malformed payload under a *valid* envelope surfaces as Avro's own `AvroRuntimeException` (not `IOException`), so it previously escaped **context-free** — the production SR path debugged worse than the static path.
  - Add `hexPreview` to the static-decode, too-short, wrong-magic, and envelope-decode error messages.

## Proven, not asserted
`malformedPayloadUnderValidEnvelopeIsWrappedWithContext` **fails when the catch is narrowed back to `IOException`-only** (the `AvroRuntimeException` leaks), passes with the widening — I verified both directions. `DiagnosticsTest` pins the shared `hexPreview` contract (empty / short / exactly-limit / truncated); the wrong-magic and too-short tests now assert the hex preview appears.

## Verification
`./gradlew :lib:kpipe-core:test :lib:kpipe-format-json:test :lib:kpipe-format-avro:test :lib:kpipe-format-protobuf:test` — BUILD SUCCESSFUL. No behavioral change to JSON/Protobuf (identical hex output); Avro error messages gain byte context.


eschizoid:
**Round-1 review update:** the type/API lens correctly caught that `registry.PipelineDiagnostics` already owned a byte-preview (`hexDump`), so the first cut (a new `diagnostics.Diagnostics` in a fresh exported package) re-introduced the very divergence the hoist was meant to remove. Reconciled in `9f6b877`:

- Single home: `registry.WireDiagnostics` (already-exported package) — the new `diagnostics` package and its `exports` line are removed, so the JPMS surface *shrank*.
- One renderer, two documented sibling presentations: `hexPreview` (compact, 8-byte, ` ...`) for embedding in a format message; `hexDump(data, maxBytes)` (fuller, `…(N bytes total)`) for the pipeline boundary. `PipelineDiagnostics.hexDump` now delegates to it — **output byte-identical** (verified by diffing over lengths 0–40 and by its existing tests).
- Renamed to disambiguate from `PipelineDiagnostics`; added a resolver-throws-propagates-unwrapped Avro test hardening the #244 boundary.

Fleet converged MERGEABLE across correctness, type/API (after this fix), and doc-accuracy.

PR #331, opened 2026-09-14
docs: point the registry's conditional processing at TypedPipelineBuilder.when

Zir0-93:
`docs/ESCAPE-HATCHES.md` [line 200](https://github.com/eschizoid/kpipe/blob/b49db7105f89580ac5e0f7d216f8c68a2e7827cb/docs/ESCAPE-HATCHES.md?plain=1#L200) says conditional processing is `Operators.when(predicate, ifTrue, ifFalse)`, but `Operators` has no `when` method, so the snippet doesn't compile. That section covers the explicit registry surface, where the method is [`TypedPipelineBuilder.when(...)`](https://github.com/eschizoid/kpipe/blob/b49db7105f89580ac5e0f7d216f8c68a2e7827cb/lib/kpipe-core/src/main/java/io/github/eschizoid/kpipe/registry/TypedPipelineBuilder.java#L66), reached through `registry.pipeline(format)`. ([`Stream.when(...)`](https://github.com/eschizoid/kpipe/blob/b49db7105f89580ac5e0f7d216f8c68a2e7827cb/lib/kpipe-api/src/main/java/io/github/eschizoid/kpipe/Stream.java#L81) is the fluent-API equivalent.)

One-line docs change.

Here's the check that flagged this, run on a fork of kpipe: [check run](https://github.com/hadi-technology/kpipe/pull/1/checks?check_run_id=104154663755).


eschizoid:
@Zir0-93 Thank you!

PR #340, opened 2026-09-18
test(consumer): gate the dispatch counter's increment-under-lock placement with Fray

### LINKED ISSUE #338: test(consumer): gate the batch dispatch counter's increment-under-lock placement with Fray
`BatchPipelineWrapper` increments its in-flight dispatch counter while the flush lock is still held, rather than when the dispatch starts. That placement is what lets `close()` see work outstanding when a tick has released the lock but has not yet begun dispatching — narrow window, and the one the drain exists to cover.

Nothing gates it. `BatchCloseDrainsDispatchTest` calls `close()` only after a callback has already fired, so it never enters the unlock-to-dispatch-start window and passes with the increment in either position. Moving the increment into the dispatch reintroduces a narrower version of the race with the suite green. The field's javadoc says so outright today, which is honest but is a disclaimer rather than a guard.

## Why Fray rather than threads

The question is whether an interleaving exists where `close()` observes zero, which is what Fray explores systematically and ordinary threads only sample. It is already wired here — `frayTest`, `@FrayTest(iterations = 500)` — and `BatchWrapperLockFrayTest` already drives this class.

## Sketch

Respecting the constraint that test documents (a live scheduler thread stops Fray finishing an iteration, so no age tick): pre-load `maxSize - 1` records, then race one `enqueue` that trips the flush against a `close()`. Have the callback increment and decrement its own `AtomicInteger`; the closing thread reads it immediately after `close()` returns and asserts zero — a single atomic read, no two-flag race.

With the increment under the lock every interleaving should be safe by construction. With it moved into the Runnable, Fray should schedule the closer between `lock.unlock()` and the Runnable's first statement, the closer sees zero, and the assertion fires.

## Status

Unvalidated. The box this was designed on reports `You are building on an unsupported OS/architecture combination; Fray JVMTI agent will not be added as a dependency` (Darwin x86_64), and `FrayInstrumentationSelfCheckTest` exists to make that state red rather than let it read green — so the sketch has never been executed. Validate it where Fray actually instruments, and keep the javadoc disclaimer until it is green there. An unvalidated gate is worse than an accurate disclaimer.

Found while reviewing #337.

eschizoid:
Closes #338.

`close()` waits on a count of dispatches still outstanding, and that count is incremented **while the flush lock is still held** rather than when the dispatch begins. That placement is what lets a `close()` arriving after a tick released the lock — but before its dispatch started — still see the work. Moving the increment into the dispatch narrows the window rather than closing it.

Nothing gated it. `BatchCloseDrainsDispatchTest` calls `close()` only after a callback has already fired, so it never enters the unlock-to-dispatch-start window and passes with the increment in either position. The window is a handful of instructions, which is exactly the case Fray schedules deliberately instead of hoping a racing test lands in it.

## The assertion is an ordering one, and that matters

The design sketched on the issue was to sample a live-callback counter the instant `close()` returns and assert zero. That gives a false pass: in the broken interleaving `close()` slips past **before** the dispatch starts, so nothing is running yet and the sample reads clean. The sample looking clean *is* the bug.

What distinguishes the two placements is whether any callback runs **after** `close()` has returned. So the closing thread sets a flag on return, and the callbacks check it — a violation is recorded only if a callback runs on the wrong side of that line.

## Validation

**Fray cannot run on the machine this was written on** — Darwin x86_64, where the plugin skips the JVMTI agent and `FrayInstrumentationSelfCheckTest` fails the job rather than letting an un-instrumented suite read green. CI's `fray` job runs on `ubuntu-latest` (linux-x86_64, a supported platform), so this PR's CI is the first place the test actually executes.

The mutation check — that the gate fails with the increment moved into the dispatch — is therefore run on CI too, on a throwaway branch via `workflow_dispatch`, rather than claimed from local reasoning. Result is posted in a follow-up comment; if it does not discriminate, this PR should not merge and the javadoc disclaimer should go back.

## Javadoc

`dispatchesInFlight`'s javadoc previously said outright that no test covered the placement. It now points at this test instead. That change is only correct if CI proves the gate discriminates, so it stands or falls with the mutation result above.

## Local verification caveat

`spotlessCheck` could not be run locally: `NODE_OPTIONS` on this machine references a deleted `restore-node-options.cjs`, and node v26.4.0 cannot resolve the prettier plugin directory import that prettier-plugin-java 2.8.1 relies on. Both are environmental and neither touches this repo. CI's spotless run is the verdict on formatting here.


eschizoid:
## Mutation check: the gate discriminates

Fray does not instrument on the machine this was written on (Darwin x86_64), so both arms were run on CI, where the `fray` job executes on `ubuntu-latest`.

| arm | `fray` job |
| --- | --- |
| increment under the flush lock (this PR) | **success** |
| increment moved into the dispatch | **failure** |

The mutant was run on a throwaway branch via `workflow_dispatch` ([run 35333671776](https://github.com/eschizoid/kpipe/actions/runs/35333671776)) and the branch has been deleted. So the test fails for the reason it exists and not merely alongside the fix, which is what the javadoc change in this PR depends on.

eschizoid:
## Re-validated against the updated test

The non-vacuity assertion changed the test, so the earlier mutation result no longer covered what is on the branch. Re-run ([35336324135](https://github.com/eschizoid/kpipe/actions/runs/35336324135)):

| arm | `build` | `fray` |
| --- | --- | --- |
| increment under the flush lock (this PR) | success | **success** |
| increment moved into the dispatch | success | **failure** |

Cleaner than the first pass, where a formatting failure in `build` sat alongside the `fray` failure. Here `build` is green in both arms, so the only thing separating them is the gate.

